PR #3, opened 2020-09-07
fix(sample): correct sample

georgettica:
now sample resembles what needs to run on the cluster

also:
- shorten service name so deploy will work


RiRa12621 (approved):
/lgtm

PR #5, opened 2020-09-10
feat(routemonitor): routemonitor populate

georgettica:
this now does all the creation until the prometheus

- add route extraction
- fix params


RiRa12621 (approved):
/lgtm
/approve
Let's go!

PR #9, opened 2020-10-07
controllers/routemonitor: add initial fuzz test

RiRa12621:
Add Fuzz test for routeMonitorName and routeMonitorNamespace
That's user controlled input.
Arguably routeMonitorNamespace is double checking because Openshift should have sanitized this already


georgettica (commented):
/lgtm
/approve

PR #13, opened 2020-10-19
fix(logging): changing log level works

georgettica:


RiRa12621 (approved):
/lgtm

georgettica:
Darn! I think our image we use is blocked :(

PR #14, opened 2020-10-26
Add ClusterUrlMontitor

NautiluX:


clusterurlmonitor.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: monitoring.openshift.io/v1alpha1
```
georgettica, line 1: and this?

config/samples/monitoring_v1alpha1_clusterurlmonitor.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: monitoring.openshift.io/v1alpha1
+kind: ClusterUrlMonitor
+metadata:
+  name: clusterurlmonitor-sample
+spec:
+  # Add fields here
+  foo: bar
```
georgettica, line 7: maybe modify this file to express an actual ClusterUrlMonitor

controllers/clusterurlmonitor/clusterurlmonitor.go
```diff
@@ -0,0 +1,123 @@
...
+
+// ClusterUrlMonitorReconciler reconciles a ClusterUrlMonitor object
+type ClusterUrlMonitorReconciler struct {
+	client.Client
+	Log    logr.Logger
+	Scheme *runtime.Scheme
+}
+
+const (
+	FinalizerKey string = "clusterurlmonitor.monitoring.openshift.io/clusterurlmonitorcontroller"
```
georgettica, line 43: there is a const package for FinalizerNames

controllers/clusterurlmonitor/clusterurlmonitor.go
```diff
@@ -0,0 +1,123 @@
...
+)
+
+// +kubebuilder:rbac:groups=monitoring.openshift.io,resources=clusterurlmonitors,verbs=get;list;watch;create;update;patch;delete
+// +kubebuilder:rbac:groups=monitoring.openshift.io,resources=clusterurlmonitors/status,verbs=get;update;patch
+
+func (r *ClusterUrlMonitorReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
+	ctx := context.Background()
+	log := r.Log.WithValues("clusterurlmonitor", req.NamespacedName)
+
+	// your logic here
```
georgettica, line 53: remove stub comment

controllers/clusterurlmonitor/clusterurlmonitor.go
```diff
@@ -0,0 +1,123 @@
...
+	clusterUrlMonitor, res, err := r.GetClusterUrlMonitor(ctx, req)
+	if err != nil {
+		return utilreconcile.RequeueWith(err)
+	}
+
+	if res.ShouldStop() {
+		return utilreconcile.Stop()
+	}
+
+	blackboxExporter := blackboxexporter.NewBlackboxExporter(r.Client, log, ctx)
```
georgettica, line 63: this should be done via interface and pushed through the main.go before running

georgettica: context won't be easily passed, but this is a fight for another day

controllers/clusterurlmonitor/clusterurlmonitor.go
```diff
@@ -0,0 +1,123 @@
...
+	s := NewSupplement(clusterUrlMonitor, r.Client, r.Log, blackboxExporter)
```
georgettica, line 64: same here

controllers/clusterurlmonitor/clusterurlmonitor.go
georgettica, line 63: this creates the resource before testing if it's marked for deletion.. do the tests you ran on it cover that scenario?
creation + deletion?

controllers/clusterurlmonitor/clusterurlmonitor.go
```diff
@@ -0,0 +1,123 @@
...
+	if err != nil {
+		return utilreconcile.RequeueWith(err)
+	}
+
+	serviceMonitor, err := s.EnsureServiceMonitor()
+	if err != nil {
+		return utilreconcile.RequeueWith(err)
+	}
+
+	res, err = s.EnsureDeletionProcessed(serviceMonitor)
```
georgettica, line 84: this doesn't make sense.. when a delete is requested it should be handled before all the creation code

controllers/clusterurlmonitor/clusterurlmonitor.go
```diff
@@ -0,0 +1,123 @@
...
+}
+
+func (r *ClusterUrlMonitorReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	return ctrl.NewControllerManagedBy(mgr).
+		For(&monitoringv1alpha1.ClusterUrlMonitor{}).
+		Complete(r)
+}
+
+// GetClusterUrlMonitor return the ClusterUrlMonitor that is tested
+func (r *ClusterUrlMonitorReconciler) GetClusterUrlMonitor(ctx context.Context, req ctrl.Request) (v1alpha1.ClusterUrlMonitor, utilreconcile.Result, error) {
```
georgettica, line 102: should be seperated to a supplement file?

controllers/clusterurlmonitor/clusterurlmonitor_supplement.go
```diff
@@ -0,0 +1,99 @@
...
+	utilfinalizer "github.com/openshift/route-monitor-operator/pkg/util/finalizer"
+	"github.com/openshift/route-monitor-operator/pkg/util/reconcile"
+	"github.com/openshift/route-monitor-operator/pkg/util/templates"
+	monitoringv1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1"
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+type ClusterUrlMonitorSupplement struct {
```
georgettica, line 20: if you created a different struct, you don't need to be in the same package as the reconciler, and the supplement can be used as a package

controllers/clusterurlmonitor/clusterurlmonitor_supplement.go
```diff
@@ -0,0 +1,99 @@
...
+	log               logr.Logger
+	ctx               context.Context
+	blackboxExporter  *blackboxexporter.BlackboxExporter
+}
+
+func NewSupplement(clusterUrlMonitor v1alpha1.ClusterUrlMonitor, client client.Client, log logr.Logger, blackboxExporter *blackboxexporter.BlackboxExporter) *ClusterUrlMonitorSupplement {
+	return &ClusterUrlMonitorSupplement{clusterUrlMonitor, client, log, context.Background(), blackboxExporter}
+}
+
+func (s *ClusterUrlMonitorSupplement) EnsureFinalizer() (reconcile.Result, error) {
```
georgettica, line 32: this is also common logic (besides the finalizer key)

should this be refactored

controllers/clusterurlmonitor/clusterurlmonitor_supplement.go
```diff
@@ -0,0 +1,99 @@
...
+	if !utilfinalizer.Contains(s.clusterUrlMonitor.GetFinalizers(), FinalizerKey) {
+		utilfinalizer.Add(&s.clusterUrlMonitor, FinalizerKey)
+		err := s.client.Update(s.ctx, &s.clusterUrlMonitor)
+		return reconcile.RequeueReconcileWith(err)
+	}
+	return reconcile.ContinueReconcile()
+}
+
+func (s *ClusterUrlMonitorSupplement) EnsureServiceMonitor() (*monitoringv1.ServiceMonitor, error) {
```
georgettica, line 41: I feel this function name is cut in the middle "ensureservicemonitor<SOMETHING>" but that may be because I am used to reading it

controllers/clusterurlmonitor/clusterurlmonitor_supplement.go
```diff
@@ -0,0 +1,99 @@
...
+	namespacedName := templates.TemplateForServiceMonitorName(s.clusterUrlMonitor.Namespace, s.clusterUrlMonitor.Name)
+	serviceMonitor := &monitoringv1.ServiceMonitor{}
+
+	clusterDomain, err := s.GetClusterDomain(s.ctx)
+	if err != nil {
+		return serviceMonitor, err
+	}
+	clusterUrl := s.clusterUrlMonitor.Spec.Prefix + clusterDomain + ":" + s.clusterUrlMonitor.Spec.Port + s.clusterUrlMonitor.Spec.Suffix
```
georgettica, line 49: this is not a full solution, so commiting as is won't work, but it gets the point across
```suggestion
         import fmt
	clusterUrl :=fmt.Print("%s%s:%s%s", s.clusterUrlMonitor.Spec.Prefix , clusterDomain ,s.clusterUrlMonitor.Spec.Port , s.clusterUrlMonitor.Spec.Suffix)
```

controllers/clusterurlmonitor/clusterurlmonitor_supplement.go
```diff
@@ -0,0 +1,99 @@
...
+
+	// Does the resource already exist?
+	if err = s.client.Get(s.ctx, namespacedName, serviceMonitor); err != nil {
+		// If this is an unknown error
+		if !k8serrors.IsNotFound(err) {
+			// return unexpectedly
+			return serviceMonitor, err
```
georgettica, line 56: why is servicemonitor returned in this case? this can send the wrong message back to the user

controllers/dummy_controller.go
```diff
@@ -0,0 +1,6 @@
+package controllers
```
georgettica, line 1: interesting...

controllers/routemonitor/routemonitor_supplement.go
```diff
@@ -4,34 +4,34 @@ import (
...
-	}
-	// Creating Service after because:
-	//
-	// A Service should not point to an empty target (Deployment)
-	if err := r.EnsureBlackBoxExporterServiceExists(ctx); err != nil {
-		return err
-	}
-	return nil
+//go:generate mockgen -source $GOFILE -destination ../../pkg/util/test/generated/mocks/$GOPACKAGE/blackboxexporter.go -package $GOPACKAGE BlackboxExporter
+type BlackboxExporter interface {
```
georgettica, line 13: why isn't that in the interface file?

controllers/routemonitor/routemonitor_test.go
```diff
@@ -38,22 +38,22 @@ var _ = Describe("Routemonitor", func() {
...
-		ensureBlackBoxExporterDeploymentAbsent helper.MockHelper
-		ensureBlackBoxExporterDeploymentExists helper.MockHelper
-		ensureBlackBoxExporterServiceExists    helper.MockHelper
-		ensureFinalizerAbsent                  helper.MockHelper // utilreconcile.Result
+		update                                helper.MockHelper
+		delete                                helper.MockHelper
+		get                                   helper.MockHelper
+		create                                helper.MockHelper
+		ensureServiceMonitorResourceAbsent    helper.MockHelper
+		shouldDeleteBlackBoxExporterResources helper.MockHelper //blackbox.ShouldDeleteBlackBoxExporter
```
georgettica, line 49:
```suggestion
		shouldDeleteBlackBoxExporterResources helper.MockHelper
```

controllers/routemonitor/routemonitor_test.go
```diff
@@ -38,22 +38,22 @@ var _ = Describe("Routemonitor", func() {
...
+		ensureBlackBoxExporterResourcesAbsent helper.MockHelper
+		ensureBlackBoxExporterResourcesExist  helper.MockHelper
+		ensureFinalizerAbsent                 helper.MockHelper // utilreconcile.Result
```
georgettica, line 52:
```suggestion
		ensureFinalizerAbsent                 helper.MockHelper
```

georgettica (commented):
didn't go through it all.. but got some stuff I wanted to be fixed

controllers/routemonitor/routemonitor.go
```diff
@@ -63,8 +64,10 @@ func (r *RouteMonitorReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error
 	shouldDelete := routeMonitor.WasDeleteRequested()
 	log.V(2).Info("Response of WasDeleteRequested", "shouldDelete", shouldDelete)
 
+	blackboxExporter := blackboxexporter.NewBlackboxExporter(r.Client, log, ctx)
```
georgettica, line 67: I think this should be injected from the main and not here

controllers/routemonitor/routemonitor_supplement.go
```diff
@@ -4,34 +4,34 @@ import (
...
+	EnsureBlackBoxExporterResourcesExist() error
+	EnsureBlackBoxExporterResourcesAbsent() (reconcile.Result, error)
+	ShouldDeleteBlackBoxExporterResources() (blackbox.ShouldDeleteBlackBoxExporter, error)
+}
+
+func (r *RouteMonitorReconciler) EnsureBlackBoxExporterResourcesExists(ctx context.Context, blackboxExporter BlackboxExporter) error {
+	return blackboxExporter.EnsureBlackBoxExporterResourcesExist()
```
georgettica, line 20: if this just wraps the call, move it to use the actual struct and mock it with an interface

controllers/routemonitor/routemonitor_test.go
```diff
@@ -161,6 +151,7 @@ var _ = Describe("Routemonitor", func() {
 	Describe("EnsureRouteMonitorAndDependenciesAbsent", func() {
 		BeforeEach(func() {
 			// Arrange
+			ensureBlackBoxExporterResourcesAbsentResponse = utilreconcile.Result{Continue: true}
```
georgettica, line 154: I used the utilreconcile to create a `ContinueOperation`

go.mod
```diff
@@ -8,7 +8,7 @@ require (
 	github.com/google/gofuzz v1.2.0
 	github.com/onsi/ginkgo v1.12.1
 	github.com/onsi/gomega v1.10.1
-	github.com/openshift/api v3.9.0+incompatible
+	github.com/openshift/api v0.0.0-20200116145750-0e2ff1e215dd
```
georgettica, line 11: why did that version change?

pkg/blackboxexporter/blackboxexporter.go
```diff
@@ -0,0 +1,243 @@
+package blackboxexporter
+
+import (
+	"context"
+
+	"github.com/go-logr/logr"
+	"github.com/openshift/route-monitor-operator/api/v1alpha1"
```
georgettica, line 7: nit: I would have seperated the imports to logical grouping-

what is local (inside route-monitor-operator)
external base packages "context"
and third party as I see fit (this can be split into more but that is already up to you

pkg/blackboxexporter/blackboxexporter.go
```diff
@@ -0,0 +1,243 @@
...
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/util/intstr"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	//api's used
+
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	//local packages
```
georgettica, line 19: this is where I had the local packages in, if you don't want it you can remove the comment

pkg/blackboxexporter/blackboxexporter.go
```diff
@@ -0,0 +1,243 @@
...
+)
+
+type BlackboxExporter struct {
+	Client client.Client
+	Log    logr.Logger
+	Ctx    context.Context
+}
+
+func NewBlackboxExporter(client client.Client, log logr.Logger, ctx context.Context) *BlackboxExporter {
```
georgettica, line 28: this doesn't fit what I did in the other packages.. do you want to fix the others or fix this one?

what I had was `func New(...)` which would make using it look like:

```
b := blackboxexporter.New(...)
```

pkg/blackboxexporter/blackboxexporter.go
```diff
@@ -0,0 +1,243 @@
+package blackboxexporter
+
+import (
+	"context"
+
+	"github.com/go-logr/logr"
+	"github.com/openshift/route-monitor-operator/api/v1alpha1"
+	"github.com/openshift/route-monitor-operator/pkg/const/blackbox"
```
georgettica, line 8: maybe change the package name in the import as they are very similar by name but mean something else by context..

maybe constblackboxexporter?

pkg/blackboxexporter/blackboxexporter.go
```diff
@@ -0,0 +1,243 @@
...
+)
+
+type BlackboxExporter struct {
+	Client client.Client
+	Log    logr.Logger
+	Ctx    context.Context
+}
+
+func NewBlackboxExporter(client client.Client, log logr.Logger, ctx context.Context) *BlackboxExporter {
+	return &BlackboxExporter{client, log, ctx}
```
georgettica, line 29: didn't know you could do that... does the order of these matter?

pkg/blackboxexporter/blackboxexporter.go
```diff
@@ -0,0 +1,243 @@
...
+	//api's used
+
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	//local packages
+)
+
+type BlackboxExporter struct {
+	Client client.Client
```
georgettica, line 23: if you would have changed the code to look like
```
client.Client
```
you could just use the struct like:
```
b := BlackboxExporter{}
b.Client.List()
b.List() // this would also work
```

Makefile
```diff
@@ -35,7 +35,7 @@ manager: generate fmt vet
 
 # Run against the configured Kubernetes cluster in ~/.kube/config
 run: generate fmt vet manifests
-	go run ./main.go
+	go run ./main.go --zap-log-level=5
```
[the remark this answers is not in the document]
georgettica: also @NautiluX , maybe adding an entrypoint which looks like:

```
run-verbose: generate fmt vet manifests
  go run ./main.go --zap-log-level=5
```
and leave this one as is?

we want to test locally the same way we ship the operator..
living in a raised debug level should not be our default mode of operation IMO

pkg/blackboxexporter/blackboxexporter.go
```diff
@@ -0,0 +1,243 @@
...
+	if err := b.Client.List(b.Ctx, routeMonitors); err != nil {
+		return blackbox.KeepBlackBoxExporter, err
+	}
+
+	clusterUrlMonitors := &v1alpha1.ClusterUrlMonitorList{}
+	if err := b.Client.List(b.Ctx, clusterUrlMonitors); err != nil {
+		return blackbox.KeepBlackBoxExporter, err
+	}
+
+	if len(clusterUrlMonitors.Items)+len(routeMonitors.Items) > 1 {
```
georgettica, line 44: this means that if we passed that if we have three scenarios:
1. there is 1 routeMonitor and no clusterUrlMonitor
2. there is 1 clusterUrlMonitor and no routeMonitor
3. there are none of both

wouldn't that be clearer to write in if statements then range on the items?

pkg/util/templates/templates.go
```diff
@@ -0,0 +1,68 @@
+package templates
```
georgettica, line 1: change to singular form
```suggestion
package template
```

georgettica (commented):
one nit, finished reading it all..

altogether seems solid work, but still needs to be flushed out / flush prev work to make the codebase look uniform

api/v1alpha1/clusterurlmonitor_types.go
```diff
@@ -0,0 +1,66 @@
...
+
+	// Foo is an example field of ClusterUrlMonitor. Edit ClusterUrlMonitor_types.go to remove/update
+	Prefix string `json:"prefix,omitempty"`
+	Suffix string `json:"suffix,omitempty"`
+	Port   string `json:"port,omitempty"`
+}
+
+// ClusterUrlMonitorStatus defines the observed state of ClusterUrlMonitor
+type ClusterUrlMonitorStatus struct {
+	// INSERT ADDITIONAL STATUS FIELD - define observed state of cluster
```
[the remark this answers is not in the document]
NautiluX: this would be a unnecessary redundancy. If anything, I would add a reference to the `ServiceMonitor` we created. You think that would be helpful?

controllers/clusterurlmonitor/clusterurlmonitor.go
NautiluX: What's the advantage of having all the constants in a single package? I think this makes it less flexible as you always need to look into a different file (even package) to figure out the value of a constant, and when moving a package you always need to carve out the constants of the constant package and move them along with it.

controllers/clusterurlmonitor/clusterurlmonitor_supplement.go
NautiluX: what's the advantage of moving a single struct into its own package? As long as the only user is the `clusterurlmonitor` package I would prefer to keep it there.

controllers/dummy_controller.go
NautiluX: Not sure if there is a better way to do it. I guess the sdk just assumes you don't love around controllers after the generation. I also think putting the controller back and keeping the supplements in the package where they are currently would make sense. That way the api is separated from the business logic. But that maybe a future improvement, for now I wanted to make the new package structure for the existing one.

controllers/routemonitor/routemonitor_supplement.go
NautiluX: See comment about constant. Interfaces should stay with the user.

go.mod
NautiluX: the I ingress was not in config/v1 version 3.9

pkg/blackboxexporter/blackboxexporter.go
NautiluX: Uhm I think those imports are ordered by go imports. Will have another look

pkg/blackboxexporter/blackboxexporter.go
NautiluX: Yes

pkg/blackboxexporter/blackboxexporter.go
NautiluX: Actually I like that the logic is separated because I can easily see which functions I use from the client, and  which belong to the struct itself. I use to think about inherenting other structs as a "is-a" relationship, vs a "has-a" relationship with a named member. Is one of the primary functions of this struct to act as a kubernetes client? No, so I'd separate it.

pkg/util/templates/templates.go
NautiluX: I don't have strong feelings about this, but since there is not one template in the package but it collects all commonly used templates, it feels right to give it a plural name.

api/v1alpha1/clusterurlmonitor_types.go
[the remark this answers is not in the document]
georgettica: ok, if this is the way then you can update the RouteMonitor that uses that way

what I was using it there is to seperate the RouteURL creation and the injection to the servicemonitor

controllers/dummy_controller.go
georgettica: will raise the Question to the operator folks to see there ideas (maybe something we didn't configure 100%)

controllers/routemonitor/routemonitor_supplement.go
georgettica: oh, so you are saying the interface file I did there should be merged with the supplement?
I am totally fine with that

go.mod
georgettica: yeah, also it broke for everyone and now they use what you did

see https://github.com/openshift/api/issues/456 to track

pkg/blackboxexporter/blackboxexporter.go
georgettica: when you add a newline between imports, they are sorted inside that section

pkg/blackboxexporter/blackboxexporter.go
georgettica: oki, just wanted to use the shorthand.. can refactor to consolidate to your view.

I would like that this discussion to sit in a comment here / in a coding guide as this is a fine line I think we should keep note of when composing together interfaces / structs

api/v1alpha1/clusterurlmonitor_types.go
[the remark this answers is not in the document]
NautiluX: I think we could adjust that later. For now the `ServiceMonitor` is linked to the CR just by its name. However, there may be nameclashes between `RouteMonitor` and `ClusterUrlMonitor` generated `ServiceMonitors`, if both have the same name.

Makefile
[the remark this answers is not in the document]
NautiluX: agreed

clusterurlmonitor-2.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: monitoring.openshift.io/v1alpha1
+kind: ClusterUrlMonitor
+metadata:
+  name: clusterurlmonitor-sample-2
```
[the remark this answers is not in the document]
NautiluX: they should not be here.

clusterurlmonitor-2.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: monitoring.openshift.io/v1alpha1
```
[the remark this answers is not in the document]
NautiluX: good question! It should not

clusterurlmonitor.yaml
NautiluX: :recycle:

controllers/clusterurlmonitor/clusterurlmonitor.go
NautiluX: yep

controllers/clusterurlmonitor/clusterurlmonitor.go
NautiluX: the order shouldn't matter, in the worst case we would create resources right before we delete it in the unlikely case someone creates a CR that already has a deletion timestamp. However to match how other operators work, I'll move it to the top.

api/v1alpha1/clusterurlmonitor_types.go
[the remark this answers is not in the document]
georgettica: so then the easiest way to solve this possible bug is to add the creating resource to the name (`routemonitor-name-namespace`)

pkg/util/templates/templates.go
georgettica: we can resolve this in that case, we might circle back to that sometime

georgettica:
saw a `review requested` tag to I went through it now, but I don't see that all the notes were addressed

controllers/clusterurlmonitor/clusterurlmonitor.go
NautiluX: yes

controllers/clusterurlmonitor/clusterurlmonitor_supplement.go
NautiluX: how would you call the function?

controllers/clusterurlmonitor/clusterurlmonitor_supplement.go
NautiluX: Why is this better than concatenating?

controllers/clusterurlmonitor/clusterurlmonitor_supplement.go
NautiluX: no it won't send wrong messages back, but it's now not returned anymore due to a refactoring I did :)

controllers/routemonitor/routemonitor_test.go
NautiluX: :shrug:

controllers/routemonitor/routemonitor_test.go
NautiluX: :shrug:

controllers/routemonitor/routemonitor_supplement.go
NautiluX: ack

controllers/routemonitor/routemonitor_test.go
NautiluX: ack

pkg/blackboxexporter/blackboxexporter.go
NautiluX: since you can't get imports wrong too easy in go, I think it's not worth the effort to order them, but I'll give it a try :D

pkg/blackboxexporter/blackboxexporter.go
NautiluX: sure

pkg/blackboxexporter/blackboxexporter.go
NautiluX: What do you think about moving them to the blackboxexporter package?

pkg/blackboxexporter/blackboxexporter.go
NautiluX: It's a matter of code style not functionality, so I don't see a reason to comment that each time I don't do it :D

pkg/blackboxexporter/blackboxexporter.go
NautiluX: You don't think it's clear that we do not want to remove if both lists together contain more than 1 item?

controllers/clusterurlmonitor/clusterurlmonitor.go
NautiluX: see below, moved deletion

config/samples/monitoring_v1alpha1_clusterurlmonitor.yaml
NautiluX: didn't know I needed to do this by hand

controllers/clusterurlmonitor/clusterurlmonitor.go
NautiluX: did that, although I won't touch main.go for it but rather wrap the Reconcile function.

controllers/clusterurlmonitor/clusterurlmonitor.go
NautiluX: same here

controllers/clusterurlmonitor/clusterurlmonitor_supplement.go
NautiluX: Not sure if it's worth the effort given it'll need to be agnostic of the type. In the end I think I would wrap that common functionality again in a function here.

controllers/routemonitor/routemonitor.go
NautiluX: done. Although I'd think wrapping Reconcile would be easier as it would reduce the lifetime of objects dramatically.

pkg/blackboxexporter/blackboxexporter.go
georgettica: what I mean is that the steps after that were not changed..

testing if ALL the resources are not deleting makes it seem like there is more than one.

Makefile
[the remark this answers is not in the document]
RiRa12621: conversation was resolved, but i don't see it changed back

controllers/routemonitor/routemonitor_supplement.go
```diff
@@ -4,40 +4,32 @@ import (
...
+	EnsureBlackBoxExporterResourcesExist() error
+	EnsureBlackBoxExporterResourcesAbsent() error
+	ShouldDeleteBlackBoxExporterResources() (blackbox.ShouldDeleteBlackBoxExporter, error)
 }
 
 func (r *RouteMonitorReconciler) EnsureRouteMonitorAndDependenciesAbsent(ctx context.Context, routeMonitor v1alpha1.RouteMonitor) (utilreconcile.Result, error) {
 	log := r.Log.WithName("Delete")
 
-	shouldDeleteBlackBoxResources, err := r.ShouldDeleteBlackBoxExporterResources(ctx, routeMonitor)
+	shouldDeleteBlackBoxResources, err := r.BlackboxExporter.ShouldDeleteBlackBoxExporterResources()
```
georgettica, line 21: it's used here but not initialized in the Reconcile, might cause issues

georgettica (commented):
did some file scanning. will pull to my machine and test there for the changes :/

RiRa12621:
/hold 
until all comments are adressed and we found some general directives on the way forward in terms of code style and the likes

controllers/clusterurlmonitor/clusterurlmonitor.go
NautiluX: it's a different one, you think it should be the same string?

pkg/blackboxexporter/blackboxexporter.go
NautiluX: did that, WDYT?

pkg/blackboxexporter/blackboxexporter.go
NautiluX: refactored that function.

controllers/clusterurlmonitor/clusterurlmonitor_supplement.go
```diff
@@ -0,0 +1,152 @@
...
+func (s *ClusterUrlMonitorSupplement) doesServiceMonitorExist(namespacedName types.NamespacedName) (bool, error) {
+	_, err := s.getServiceMonitor(namespacedName)
+	if k8serrors.IsNotFound(err) {
+		return false, nil
+	}
+	return true, err
+}
+
+// GetClusterUrlMonitor return the ClusterUrlMonitor that is tested
+func GetClusterUrlMonitor(client client.Client, log logr.Logger, req ctrl.Request, ctx context.Context) (v1alpha1.ClusterUrlMonitor, utilreconcile.Result, error) {
```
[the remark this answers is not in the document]
NautiluX: because it's used in that func, didn't want to remove those logs

controllers/clusterurlmonitor/clusterurlmonitor_test.go
```diff
@@ -0,0 +1,277 @@
...
+		m.FailReason = fmt.Sprintf("Unexpected type passed: want '%T', got '%T'", v1alpha1.ClusterUrlMonitor{}, x)
+		return false
+	}
+	m.ActualClusterUrlMonitor = *ref.DeepCopy()
+	return true
+}
+
+func (m *ClusterUrlMonitorMatcher) String() string {
+	return "Fail reason: " + m.FailReason
+}
```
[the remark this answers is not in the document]
NautiluX: it's a gomock matcher (https://godoc.org/github.com/golang/mock/gomock#Matcher)

controllers/clusterurlmonitor/clusterurlmonitor_test.go
```diff
@@ -0,0 +1,277 @@
...
+				Expect(clusterUrlMonitorMatcher.ActualClusterUrlMonitor.Finalizers).To(ContainElement(FinalizerKey))
+			})
+		})
+
+		When("the finalizer already exists", func() {
+			BeforeEach(func() {
+				clusterUrlMonitor.Finalizers = []string{clusterurlmonitor.FinalizerKey}
+			})
+			It("doesn't change the ClusterUrlMonitor", func() {
+				oldClusterUrlMonitor := clusterUrlMonitor.DeepCopy()
```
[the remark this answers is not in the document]
NautiluX: good question. Used `DeepCopy` to make sure it definitely is a copy.

controllers/clusterurlmonitor/clusterurlmonitor_test.go
```diff
@@ -0,0 +1,277 @@
...
+				ingress := configv1.Ingress{
+					Spec: configv1.IngressSpec{
+						Domain: clusterDomain,
+					},
+				}
+				gomock.InOrder(
+					mockClient.EXPECT().Get(gomock.Any(), gomock.Any(), gomock.Any()).Return(fakeNotFound),
+					mockClient.EXPECT().Get(gomock.Any(), gomock.Any(), gomock.Any()).SetArg(2, ingress),
+				)
+				mockClient.EXPECT().Create(gomock.Any(), serviceMonitorMatcher).Times(1)
```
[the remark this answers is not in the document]
NautiluX: I even thought about moving it to the It, because this is a core part of the test: Is the new object created?

controllers/clusterurlmonitor/clusterurlmonitor_test.go
```diff
@@ -0,0 +1,277 @@
...
+			It("doesn't update the ServiceMonitor", func() {
+				err := sup.EnsureServiceMonitor()
+				Expect(err).NotTo(HaveOccurred())
+			})
+		})
+	})
+
+	Describe("EnsureDeletionProcessed", func() {
+		BeforeEach(func() {
+			clusterUrlMonitor.Finalizers = []string{FinalizerKey}
```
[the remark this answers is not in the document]
NautiluX: I think I only set/use this string in this context

controllers/routemonitor/routemonitor.go
```diff
@@ -25,14 +25,16 @@ import (
...
 )
 
 // RouteMonitorReconciler reconciles a RouteMonitor object
 type RouteMonitorReconciler struct {
 	client.Client
-	Log    logr.Logger
-	Scheme *runtime.Scheme
+	Log              logr.Logger
+	Scheme           *runtime.Scheme
+	BlackboxExporter BlackboxExporter
```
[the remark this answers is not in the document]
NautiluX: for the dependency injection from `main.go`

controllers/routemonitor/routemonitor_supplement.go
NautiluX: all dependencies of `RouteMonitorReconcile` are injected from `main.go`, I'd prefer to do that in the Reconcile as well, but I didn't want to change the whole existing code structure.

api/v1alpha1/clusterurlmonitor_types.go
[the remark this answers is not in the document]
NautiluX: created card OSD-5709 to follow-up on this

controllers/routemonitor/routemonitor_supplement.go
NautiluX: I don't think we need necessarily need to change that, as it lives in a different package and is its own controller. So it can have a slightly different design.

NautiluX:
Resolved some of the comments after discussing with @georgettica

controllers/clusterurlmonitor/clusterurlmonitor_test.go
```diff
@@ -0,0 +1,277 @@
...
+						mockBlackboxExporter.EXPECT().ShouldDeleteBlackBoxExporterResources().Return(blackbox.DeleteBlackBoxExporter, nil)
+					})
+					It("removes the servicemonitor, the blackbox exporter and cleans up the finalizer", func() {
+						mockBlackboxExporter.EXPECT().EnsureBlackBoxExporterResourcesAbsent().Times(1)
+
+						res, err := sup.EnsureDeletionProcessed()
+
+						Expect(err).NotTo(HaveOccurred())
+						Expect(res).To(Equal(reconcile.StopOperation()))
+						Expect(serviceMonitorMatcher.ActualServiceMonitor.Name).To(Equal(serviceMonitor.Name))
```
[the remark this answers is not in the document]
NautiluX: done

controllers/dummy_controller.go
NautiluX: nothing to do here for now, added a additional commit. Resolving to better see what action items are still open.

controllers/clusterurlmonitor/clusterurlmonitor_test.go
```diff
@@ -0,0 +1,277 @@
...
+				err := sup.EnsureServiceMonitor()
+				expectedUrl := prefix + clusterDomain + ":" + port + suffix
+				Expect(err).NotTo(HaveOccurred())
+				Expect(serviceMonitorMatcher.ActualServiceMonitor.Spec.Endpoints[0].Params["target"][0]).To(Equal(expectedUrl))
+			})
+		})
+		When("the ServiceMonitor exists already", func() {
+			BeforeEach(func() {
+				serviceMonitor := monitoringv1.ServiceMonitor{}
+				mockClient.EXPECT().Get(gomock.Any(), gomock.Any(), gomock.Any()).SetArg(2, serviceMonitor)
```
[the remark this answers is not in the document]
NautiluX: done

controllers/clusterurlmonitor/clusterurlmonitor_test.go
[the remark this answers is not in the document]
NautiluX: changed name of `ActualServiceMonitor`

controllers/clusterurlmonitor/clusterurlmonitor_test.go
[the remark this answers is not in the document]
georgettica: ok, seems fair, resolving

controllers/clusterurlmonitor/clusterurlmonitor_test.go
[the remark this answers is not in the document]
georgettica: oki, resolving

controllers/routemonitor/routemonitor.go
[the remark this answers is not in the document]
georgettica: ack

controllers/routemonitor/routemonitor_supplement.go
georgettica: okay.. not sure how I feel about it but we can close if this is a blocker

controllers/routemonitor/routemonitor_supplement.go
georgettica: well, that can be a next step refactor.

also I think I just missed the main.go injection 😅

georgettica (commented):
mostly approved, I will go through it now

georgettica:
/unhold
all review notes are addressed

RiRa12621:
/lgtm
Go for it once you're done testing

NautiluX:
/unhold
I tested it today, and verified it works as expected.

PR #16, opened 2020-11-06
Add integration test suite and base tooling

NautiluX:


Makefile
```diff
@@ -75,9 +76,18 @@ generate: controller-gen mockgen
...
-	docker push ${IMG}
+# Build the image with podman
+podman-build: test
+	podman build . -t ${IMG}
+
+# Push the image with podman
+podman-push:
+	podman push ${IMG} --tls-verify=false
+
+test-integration: podman-build
```
georgettica, line 87: does that mean that test-integration is based on podman only?

Makefile
```diff
@@ -75,9 +76,18 @@ generate: controller-gen mockgen
...
-# Push the docker image
-docker-push:
-	docker push ${IMG}
+# Build the image with podman
+podman-build: test
+	podman build . -t ${IMG}
+
+# Push the image with podman
+podman-push:
+	podman push ${IMG} --tls-verify=false
```
georgettica, line 85: maybe instead make the binary (podman / docker) be an env variable and should be set when you run the command
binary can be called $CONTAINER_ENGINE

hack/test-integration-setup.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+
+set -euo pipefail
+
+KUBEUSER=${KUBEUSER:-kubeadmin}
+
+oc patch configs.imageregistry.operator.openshift.io/cluster --patch '{"spec":{"defaultRoute":true}}' --type=merge
+REGISTRY=$(oc get route default-route -n openshift-image-registry -o json | jq -r .spec.host)
```
georgettica, line 8: I think you can use here the jsonpath instead of `-o json | jq -r ...`

hack/test-integration-setup.sh
```diff
@@ -0,0 +1,17 @@
...
+IMAGE_NAME=route-monitor-operator
+IMAGE=$REGISTRY/openshift-monitoring/$IMAGE_NAME
+podman build . -t "$REGISTRY/openshift-monitoring/route-monitor-operator"
```
georgettica, line 11: here again, please set the CONTAINER_ENGINE to make other env's work seamlessly

hack/test-integration-setup.sh
```diff
@@ -0,0 +1,17 @@
...
+TOKEN=$(oc whoami -t)
+podman login "$REGISTRY" -u "$KUBEUSER" -p "$TOKEN" --tls-verify=false
+podman push "$IMAGE" --tls-verify=false
+oc create imagestream "$IMAGE_NAME" -n openshift-monitoring || true
```
georgettica, line 15: maybe add comment to explain the `|| true` is to allow the code to be idempotent

int/int_test.go
```diff
@@ -0,0 +1,167 @@
...
+		var (
+			clusterUrlMonitorName      string
+			clusterUrlMonitorNamespace string
+			clusterUrlMonitor          v1alpha1.ClusterUrlMonitor
+			expectedServiceMonitorName types.NamespacedName
+		)
+		BeforeEach(func() {
+			err := i.RemoveClusterUrlMonitor(clusterUrlMonitorNamespace, clusterUrlMonitorName)
+			clusterUrlMonitorName = "fake-url-monitor"
+			clusterUrlMonitorNamespace = "default"
```
georgettica, line 41: should we also create a test with the `openshift-monitoring` namespace, as some of the resources are created there aswell?

int/int_test.go
```diff
@@ -0,0 +1,167 @@
...
+			expectedServiceMonitorName = templates.TemplateForServiceMonitorName(clusterUrlMonitorNamespace, clusterUrlMonitorName)
+			Expect(err).NotTo(HaveOccurred())
```
georgettica, line 43: shouldn't this error be much higher, as the `TemplateForServiceMonitorName` doesn't return that error?

int/int_test.go
```diff
@@ -0,0 +1,167 @@
...
+				serviceMonitor, err := i.WaitForServiceMonitor(expectedServiceMonitorName, 20)
+				Expect(err).NotTo(HaveOccurred())
+
+				Expect(serviceMonitor.Name).To(Equal(expectedServiceMonitorName.Name))
+				Expect(serviceMonitor.Namespace).To(Equal(expectedServiceMonitorName.Namespace))
+
+				clusterConfig := configv1.Ingress{}
+				err = i.Client.Get(context.TODO(), types.NamespacedName{Name: "cluster"}, &clusterConfig)
+				Expect(err).NotTo(HaveOccurred())
+				expectedUrl := clusterUrlMonitor.Spec.Prefix + clusterConfig.Spec.Domain + ":" + clusterUrlMonitor.Spec.Port + clusterUrlMonitor.Spec.Suffix
```
georgettica, line 75: same comment I gave in prev review 
```suggestion
				spec := clusterUrlMonitor.Spec                          
				expectedUrl := spec.Prefix + spec.Domain + ":" + spec.Port + spec.Suffix
```

int/integration.go
```diff
@@ -0,0 +1,141 @@
...
+func NewIntegration() (*Integration, error) {
+	scheme := runtime.NewScheme()
+	utilruntime.Must(clientgoscheme.AddToScheme(scheme))
+
+	utilruntime.Must(monitoringv1alpha1.AddToScheme(scheme))
+	utilruntime.Must(monitoringv1.AddToScheme(scheme))
+	utilruntime.Must(routev1.AddToScheme(scheme))
+	utilruntime.Must(configv1.AddToScheme(scheme))
+
+	utilruntime.Must(monitoringopenshiftiov1alpha1.AddToScheme(scheme))
```
georgettica, line 41: this can be used fro
```suggestion
        import ( constinit "github.com/openshift/route-monitor-operator/pkg/consts/test/init" )
	scheme := constinit.Scheme
```

int/integration.go
```diff
@@ -0,0 +1,141 @@
...
+	mgr, err := ctrl.NewManager(ctrl.GetConfigOrDie(), ctrl.Options{
+		Scheme: scheme,
+	})
+	if err != nil {
+		return &Integration{}, err
+	}
+	client := mgr.GetClient()
+	i := Integration{client, make(chan struct{}), mgr}
+	go func(x chan struct{}) {
+		//err := mgr.Start(i.clientChan)
```
georgettica, line 51: why is this comment here?

georgettica (commented):
one Q, why 20 seconds? not questioning or stopping the PR but still

Makefile
NautiluX: yes, this is because podman supports to push to insecure registries without much reconfiguration

Makefile
georgettica: nvm, podman is better in some ways and I will need to create docker variant for it anyways

georgettica:
went through and cleaned what I though was relevant, also tried to shorten some lines but logic seems solid.

a little sad I cannot use docker for stuff (podman for macos is nonexistant)

RiRa12621:
>a little sad I cannot use docker for stuff (podman for macos is nonexistant)
I think that's actually a problem

georgettica:
So @RiRa12621 can you hold this if you feel strongly on that

RiRa12621:
not strong enough to hold tbh

hack/test-integration-setup.sh
NautiluX: done

int/int_test.go
NautiluX: I don't understand the suggestion, usually `RouteMonitors` and `ClusterUrlMonitors` are created somewhere else and the result is generated in `openshift-monitoring`

int/int_test.go
NautiluX: done

int/integration.go
NautiluX: in this case I think there are some missing in github.com/openshift/route-monitor-operator/pkg/consts/test/init

int/integration.go
NautiluX: good question.

NautiluX:
For automation on crc we need to use podman otherwise we cannot push to a unsecure registry without changing the global docker config. I'd say we can figure out a way to support docker after this PR...

hack/test-integration-setup.sh
NautiluX: refactored to have a better readable script

NautiluX:
decided to use 20 seconds as a timeout as it felt like a good time that should be achievable by a cluster but doesn't produce a too long waiting time in case of a failure. Happy for suggestions.

georgettica:
/lgtm
/approve
glad my suggestions are taken into account, and hope the docker solution will be available aswell (for compatibility)

RiRa12621:
What's wrong with the bot
hello @openshift-ci-robot can you do your thing?!

RiRa12621 (approved):
/lgtm
/approve

int/int_test.go
georgettica: never mind, it was just to see that the resource can be created in different namespaces, and not only in the default one (has some wierd permissions)

maybe create a specific namespace and test there too?

int/integration.go
georgettica: if they are missing, add them in addition to the base ones? makes it easier to distinguish between duplication and what is required

RiRa12621:
@NautiluX i think you can go ahead and merge

PR #17, opened 2020-11-06
Add readme section for ClusterUrlMonitors

NautiluX:


RiRa12621:
Can we add an example of what this might be commonly used for?

NautiluX:
> Can we add an example of what this might be commonly used for?

done

PR #19, opened 2020-11-29
Boilerplate subscribe 1 7279a26b08765f4f8e17f23fcee15c09c9694c59

georgettica:
- [ ] Remove any other 'include' lines, unless they're for things truly
  unique to your repository. (Otherwise, consider proposing them to
  boilerplate.)

- [ ] Delete any obsolete files you're no longer including.

- [ ] Make sure you are properly integrated with codecov.io:
  https://github.com/openshift/ops-sop/blob/93d100347746ce04ad552591136818f82043c648/services/codecov.md#generate-the-codecovio-token

- [ ] Make sure your prow and app-sre pipeline configurations use the
  build_root configuration and standard targets described in the README.


georgettica:
thanks for holding

2uasimojo:
Talked with Ron about this today. We need to take another swipe at updating the makefile contents to mesh with what bp provides. We'll continue working on it.

Makefile
```diff
@@ -67,13 +69,16 @@ generate: controller-gen mockgen
...
-	docker build . -t ${IMG}
-
-# Push the docker image
-docker-push:
-	docker push ${IMG}
+# covered by boilerplate XD
+# if they come back in the next version please check how to work around the issue
+#
+## Build the docker image
+#docker-build: test
```
rporres, line 76: Boilerplate expects Dockerfile in `build/Dockerfile`. You can override this by setting `OPERATOR_DOCKERFILE`

rporres (commented):
A couple of comments. Before merging this, make sure that the targets that you are now consuming from boilerplate work as you expect

Makefile
georgettica: I tried to do my due diligance and try onboarding anyways

as seen here https://github.com/openshift/boilerplate/blob/4e0dabbee031072b635a169ee3e5dc12d0bc5168/boilerplate/openshift/golang-osd-operator/update#L16
it is hardcoded, so still needs some changes there aswell

georgettica:
```
GOOS=darwin GOARCH=amd64 CGO_ENABLED=0 GOFLAGS= GOOS=linux go build -gcflags="all=-trimpath=" -asmflags="all=-trimpath=" -o build/_output/bin/ ./cmd/manager
go build github.com/openshift/route-monitor-operator/cmd/manager: build output "build/_output/bin/" already exists and is a directory
```

Makefile
rporres: 🤦

georgettica:
as we decided to go with the old way, closing PR

PR #21, opened 2020-11-29
feat(project): make repo v3 complient

georgettica:
it seems the error https://github.com/operator-framework/operator-sdk/blob/74f5a36af68489d8e3d857f2fedbffacf207474e/internal/cmd/operator-sdk/generate/internal/genutil.go#L163 requires adding this

maybe more operations are required, but I didn't look deeper

### DIFF
--- a/PROJECT
+++ b/PROJECT
@@ -1,6 +1,7 @@
 domain: openshift.io
 layout: go.kubebuilder.io/v2
 repo: github.com/openshift/route-monitor-operator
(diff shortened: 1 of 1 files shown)


georgettica:
/hold
there may be stuff that needs to happen aswell to make the migration successfull, that is all I needed for it to work

RiRa12621:
We don't care about v3 anymkfe

PR #22, opened 2020-12-10
Boilerplate bootstrap  4e0dabbee031072b635a169ee3e5dc12d0bc5168

georgettica:
second attempt just in case it works well


Makefile
```diff
@@ -1,3 +1,10 @@
+include boilerplate/generated-includes.mk
+
+# format of full image is in ./boilerplate/openshift/golang-osd-operator/standard.mk
+#IMAGE_REPOSITORY=app-interface
+
+OPERATOR_NAME=route-monitor-operator
```
rporres, line 6: The convention is to get this from the `config.go` file, e.g. https://github.com/openshift/aws-account-operator/blob/master/config/config.go

Makefile
```diff
@@ -25,10 +32,11 @@ endif
 
 all: manager
 
-TESTS=$(shell go list ./... | grep -v /int | tr '\n' ' ')
+TESTTARGETS=$(shell go list ./... | grep -v /int | tr '\n' ' ')
```
rporres, line 35: If you have to override this, we should make it overridable in boilerplate

Makefile
```diff
@@ -25,10 +32,11 @@ endif
...
+
 # Run tests
 test: generate fmt vet manifests
-	go test $(TESTS) -coverprofile cover.out
+	go test $(TESTTARGETS) -coverprofile cover.out
```
rporres, line 39: do you really need to override this? Can't you just define `TESTOPTS`?

As for the dependencies, some of those tests are passed via golangci https://github.com/openshift/boilerplate/blob/master/boilerplate/openshift/golang-osd-operator/golangci.yml or can be added if you define `GOLANGCI_OPTIONAL_CONFIG`. See https://github.com/openshift/boilerplate/tree/master/boilerplate/openshift/golang-osd-operator#linting-and-other-static-analysis-with-golangci-lint and an example in https://github.com/app-sre/deployment-validation-operator/blob/master/.golangci.yml

rporres (commented):
Please also take account the comments in https://github.com/openshift/route-monitor-operator/pull/19 :)

georgettica:
so let's fix so at least part of the process is seamless

Makefile
georgettica: that is a future me problem

Makefile
georgettica: I cannot take if from config, as it's a kustomize folder.. is there a place I can override where it takes it from?

Makefile
georgettica: yes, we have some tests we don't want to run without pre-reqs.  testopts doesn't IIUC suit my needs

georgettica:
discarding for now, but at least more progress here

Makefile
rporres: no, just override the name as you have done

PR #23, opened 2020-12-10
georgettica/app interface old way

georgettica:
- chore(appsre): copy pasta all
- chore(appsre): sed names
- feat(bundle): create bundle


georgettica:
removed WIP as I want others to 👀  and give me the 👍  that is should work

bundle/manifests/route-monitor-operator.clusterserviceversion.yaml
```diff
@@ -0,0 +1,52 @@
...
+  - operator-sdk
+  - golang
+  - monitoring
+  - sre
+  - openshift-dedicated
+  - openshift
+  - redhat
+  links:
+  - name: Route Monitor Operator
+    url: https://route-monitor-operator.domain
```
RiRa12621, line 42:
```suggestion
    url: https://github.com/openshift/route-monitor-operator/
```

config/manifests/bases/route-monitor-operator.clusterserviceversion.yaml
```diff
@@ -0,0 +1,63 @@
...
+  - operator-sdk
+  - golang
+  - monitoring
+  - sre
+  - openshift-dedicated
+  - openshift
+  - redhat
+  links:
+  - name: Route Monitor Operator
+    url: https://route-monitor-operator.domain
```
RiRa12621, line 53:
```suggestion
    url: https://github.com/openshift/route-monitor-operator/
```

RiRa12621 (approved):
/lgtm
just small suggestion to make the github repo the home page

additionally we may want to add a test in the makefile that runs the python script but that should not and does not have to be part of this PR

RiRa12621:
/assign @bdematte @rporres 
unhold when you're ready

georgettica:
thanks @RiRa12621 !

RiRa12621:
@georgettica ping me when you're fully done for LGTM again

georgettica:
For me I think it's done. But still have some things that we need to solve for the short/long term.
As discussed privately, @rporres knows the 🛣️.

But from.what I can recall, we were thinking of clearing most of if not all of `hack/generate-operator-bundle.py`

What is missing is being able to manually set the `replaces` field

Makefile
```diff
@@ -1,5 +1,11 @@
 # Current Operator version
-VERSION ?= 0.0.1
+VERSION_MAJOR=0
+VERSION_MINOR=1
+COMMIT_NUMBER=$(shell git rev-list `git rev-list --parents HEAD | egrep "^[a-f0-9]{40}$$"`..HEAD --count)
+CURRENT_COMMIT=$(shell git rev-parse --short=7 HEAD)
+OPERATOR_VERSION=$(VERSION_MAJOR).$(VERSION_MINOR).$(COMMIT_NUMBER)-$(CURRENT_COMMIT)
+
+VERSION ?= $(OPERATOR_VERSION)
```
rporres, line 8: This is only used once in `148`. Why not using there `OPERATOR_VERSION` directly?

hack/app_sre_create_image_catalog.sh
```diff
@@ -0,0 +1,115 @@
...
+git clone \
+    --branch "$BRANCH_CHANNEL" \
+    https://app:"${APP_SRE_BOT_PUSH_TOKEN}"@gitlab.cee.redhat.com/service/saas-route-monitor-operator-bundle.git \
+    "$SAAS_OPERATOR_DIR"
+
+# remove any versions more recent than deployed hash
+REMOVED_VERSIONS=""
+if [[ "$REMOVE_UNDEPLOYED" == true ]]; then
+    DEPLOYED_HASH=$(
+        curl -s "https://gitlab.cee.redhat.com/service/app-interface/raw/master/data/services/osd-operators/cicd/saas/saas-${_OPERATOR_NAME}.yaml" | \
```
rporres, line 30: Please make sure this file is going exist in app-interface or the logic won't work 😅

hack/app_sre_create_image_catalog.sh
```diff
@@ -0,0 +1,115 @@
...
+            rm -rf "${BUNDLE_DIR:?BUNDLE_DIR var not set}/$version"
+            REMOVED_VERSIONS="$version $REMOVED_VERSIONS"
+        fi
+    done
+fi
+
+# generate bundle
+PREV_VERSION=$(ls "$BUNDLE_DIR" | sort -t . -k 3 -g | tail -n 1)
+
+./hack/generate-operator-bundle.py \
```
rporres, line 56: @georgettica and I worked yesterday in an approach that didn't need this script anymore: using `make bundle` and copying the `bundle/manifests` file

hack/app_sre_pr_check.sh
```diff
@@ -0,0 +1,17 @@
...
+set -exv
+
+CURRENT_DIR=$(dirname "$0")
+
+python "$CURRENT_DIR"/validate_yaml.py "$CURRENT_DIR"/../deploy/crds
+if [ "$?" != "0" ]; then
+    exit 1
+fi
+
+BASE_IMG="aws-account-operator"
```
rporres, line 14: copy/pasta :)

hack/generate-operator-bundle.py
```diff
@@ -0,0 +1,149 @@
+#!/usr/bin/env python3
```
rporres, line 1: this file should not be needed anymore

hack/validate_yaml.py
```diff
@@ -0,0 +1,42 @@
+# Usage
```
rporres, line 1: Where is this file used?

hack/app_sre_pr_check.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
```
rporres, line 1: Is this file going to be used at all? I understand this operator will run PR checks via Prow

rporres:
> What is missing is being able to manually set the replaces field

This cannot be merged without that

bdematte:
> > What is missing is being able to manually set the replaces field
> 
> This cannot be merged without that

I am not sure what this refers to, as the script is setting the `replaces` field (with the value provided by the caller) so looks fine to me

rporres:
> I am not sure what this refers to, as the script is setting the replaces field (with the value provided by the caller) so looks fine to me

In our yesterday's tests, replaces field was being set to something like `0.<current_minor - 1>.<current_patch>-<current_sha>`

georgettica:
what @rporres means, is if you omit fully the python file, the operator on it's own set the replaces field. we can join the two together where the python script will only change the replaces field according to it's liking.

hope that clears the issue @bdematte 

also if you would like, I can show you the contents of `make generate` and what the downstream repo look like and I think you will see the divergance

hack/app_sre_pr_check.sh
georgettica: as of now the operator in app-sre is pointing to this script.

if it's in prow that's great! It's better to be tested twice then none at all.

also IIUC this is the entrypoint if the make target doesn't exist, so the file is required

hack/validate_yaml.py
georgettica: IDK, ran `cp -R aws-account-operator/hack/. route-monitor-operator/hack` and kept the results, if it's not used I see no reason to keep it

hack/app_sre_pr_check.sh
georgettica: missed one :(

hack/app_sre_create_image_catalog.sh
georgettica: I can test this code path with you if you wish

Makefile
georgettica: I wanted to keep the codebase (which was copied from operator-sdk) the same as much as I can. if we really want to change it I can live with that

hack/generate-operator-bundle.py
georgettica: As I said multiple times, not 100% true

hack/app_sre_create_image_catalog.sh
georgettica: not exactly, as you mentioned we need the `replaces` filed to be calculated as we want. so we can just override in python for now

hack/app_sre_pr_check.sh
georgettica: oh, it's PR check.. did'nt read it good the first time.. will delete

georgettica:
@rporres @bdematte I am ready for another round of reviews

hack/app_sre_create_image_catalog.sh
rporres: Just make sure it exists whenever you create the saas file in app-interface

hack/app_sre_create_image_catalog.sh
rporres: I understood you were working in the manifest generation using `make bundle`. Hence my comment

hack/generate-operator-bundle.py
rporres: I don't really understand why we were working in the `make bundle` approach then. But if you want to keep this and work on that later, I'm fine with it

hack/validate_yaml.py
rporres: I would not keep it if it's not in use

rporres (approved):
ok, so we now use the python code only to deal with the csv... not ideal but looks better. I guess it should be changed in the future but if it works for now, let's unblock this

georgettica:
@RiRa12621 it seems we have the green light for now, so:
/unhold

georgettica:
for future reference, the commands that can be used are:
```
kustomize build config/manifests/ | operator-sdk generate packagemanifests --output-dir saas-route-monitor-operator-bundle/route-monitor-operator/ --version $FIRST_VERSION
kustomize build config/manifests/ | operator-sdk generate packagemanifests --output-dir saas-route-monitor-operator-bundle/route-monitor-operator/ --from-version  $FIRST_VERSION --version $SECOND_VERSION
```

georgettica:
if @rporres approved, then it's:
/lgtm ?
or should someone else (@RiRa12621 ) do it

PR #26, opened 2020-12-14
fix(makefile): [HACK] seperate go-generate

georgettica:
in order for our code to  pass app-interface, we need to remove that generation and keep it in our heads always

this is a VERY ugly fix and I hope it's the last


georgettica:
fingers crossed

PR #30, opened 2020-12-15
feat(hack): make command use correct creds

georgettica:

PR #32, opened 2020-12-15
partial Revert "feat(hack): make command use correct creds"

RiRa12621:
Reverts openshift/route-monitor-operator#30

And adds build support via dockerized build

### DIFF
--- a/hack/app_sre_create_image_catalog.sh
+++ b/hack/app_sre_create_image_catalog.sh
@@ -16,7 +16,7 @@ docker build -f hack/pipeline.dockerfile -t pipelinebuilder:latest ./hack/
 #vol=$(generate_mac_mount $(pwd) /root/tmp)
 base_command="docker run --rm --workdir /root/tmp --interactive --volume $(pwd):/root/tmp"
 
-make bundle KUSTOMIZE="${base_command} --entrypoint kustomize pipelinebuilder:latest" OPERATOR_SDK_COMMAND="${base_command} pipelinebuilder:latest"
+make bundle KUSTOMIZE="${base_command} --entrypoint kustomize pipelinebuilder:latest" OPERATOR_SDK="${base_command} pipelinebuilder:latest"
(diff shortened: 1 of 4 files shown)


georgettica:
Just unrevert the app-sre-build-deploy section.
It was my bad then.

And it won't work 100% because of this container running as root.

And the fact that kustomize is not dockerfiled

hack/pipeline.dockerfile
```diff
@@ -0,0 +1,16 @@
+FROM registry.access.redhat.com/ubi8/ubi-minimal:latest
+
+RUN microdnf install -y golang make which
+RUN microdnf install -y git
+
+ARG BIN=operator-sdk
+COPY $BIN /usr/local/bin/operator-sdk
```
georgettica, line 7: default binary path doesn't work, and it won't work on the remote slave

hack/app_sre_create_image_catalog.sh
```diff
@@ -9,8 +9,11 @@ _OPERATOR_NAME="route-monitor-operator"
 BRANCH_CHANNEL="$1"
 QUAY_IMAGE="$2"
 
+# Build an image locally that has all tools we need
+docker build -f hack/pipeline.dockerfile -t pipelinebuilder:latest ./hack/
+
 # generate the bundle folder
-make bundle
+make bundle OPERATOR_SDK_COMMAND="docker run --rm -w /root/tmp -i -v $(pwd):/root/tmp pipelinebuilder:latest"
```
georgettica, line 16: we would also need to override the binary to kustomize with `KUSTOMIZE=`

RiRa12621: yes you're right

RiRa12621:
/lgtm
🤞

NautiluX (commented):
/lgtm

georgettica:
@NautiluX it seems like I have the power!

PR #35, opened 2020-12-16
fix(hack): change root permissions

georgettica:
this should solve this permission error


georgettica:
untested locally but should help fix the errors

georgettica:
chown binary exists, checked now

PR #36, opened 2020-12-16
georgettica/add syncset template

georgettica:
- fix(gitignore): ignore saas folders
- feat(templates): copy generation commands
- feat(template): add commands to create the template
- chore(template): run command


georgettica:
/hold
some stuff are stock and not like we would like them to be

RiRa12621:
/assign @jharrington22 
can you please make sure this is the way it is?

hack/olm-registry/olm-artifacts-template.yaml
```diff
@@ -0,0 +1,447 @@
...
+      managed.openshift.io/gitHash: ${IMAGE_TAG}
+      managed.openshift.io/gitRepoName: ${REPO_NAME}
+      managed.openshift.io/osd: 'true'
+    name: bundle-manifests
+  spec:
+    clusterDeploymentSelector:
+      matchLabels:
+        api.openshift.com/managed: 'true'
+    resourceApplyMode: Sync
+    resources:
```
jharrington22, line 29: We need to also install the namespace

hack/generate_template.py
```diff
@@ -0,0 +1,123 @@
...
+    sa_role_names = []
+    for y in yamls:
+        if y['kind'] == 'RoleBinding':
+            if 'namespace' not in y['metadata']:
+                print("RoleBinding '{}' is missing .metadata.namespace".format(y['metadata']['name']))
+            for s in y['subjects']:
+                if 'namespace' not in s:
+                    print("RoleBinding '{}' is missing .subjects[].namespace".format(y['metadata']['name']))
+                if y['roleRef']['kind'] == "Role" and s['kind'] == 'ServiceAccount' and y['metadata']['namespace'] == s['namespace']:
+                    sa_role_names.append(y['roleRef']['name'])
```
jharrington22, line 62: Is an extra role needed by this operator? They are managed by OLM otherwise

hack/olm-registry/olm-artifacts-template.yaml
```diff
@@ -0,0 +1,447 @@
...
+      metadata:
+        creationTimestamp: null
+        name: route-monitor-operator-metrics-reader
+      rules:
+      - nonResourceURLs:
+        - /metrics
+        verbs:
+        - get
+    - apiVersion: operators.coreos.com/v1alpha1
+      kind: ClusterServiceVersion
```
jharrington22, line 201: The CSV should be in the generated bundle, all the should exist in this for operator install is the subscription, catalogsource and operatorgroup

jharrington22 (changes requested):
For a operator to be deployed unless it requires extra configuration outside of OLM you should only need an OC template with 
- Namespace
- CatalogSource
- OperatorGroup
- Subscription

See the custom-domains-operator for example https://github.com/openshift/custom-domains-operator/blob/master/hack/olm-registry/olm-artifacts-template.yaml

These generally can stay static because only templated values like container versions and channels change.

hack/olm-registry/olm-artifacts-template.yaml
georgettica: We install the operator in the OpenShift monitoring namespace, so I am not sure we need to

hack/olm-registry/olm-artifacts-template.yaml
georgettica: If this is the case, where would you point me to look for a good example?
Also I took the information from cloud-ingress-operator..
Is that repo doing the generation incorrectly aswell?

hack/generate_template.py
georgettica: If they are managed by old, they should all be here

hack/olm-registry/olm-artifacts-template.yaml
georgettica: If this is the case, where would you point me to look for a good example?
Also I took the information from.

hack/generate_template.py
jharrington22: Managed by old? What does that mean?

hack/olm-registry/olm-artifacts-template.yaml
jharrington22: Ok yeah, if the namespace is always going to exist that is fine

hack/generate_template.py
RiRa12621: Typo from iOS 
Old == OLM

PR #37, opened 2020-12-17
feat(hack): root perm fixing

georgettica:


georgettica:
@RiRa12621 is seems @rporres wants us to merge it in case it works.. WDYT?

RiRa12621:
/lgtm
worth the try

georgettica:
damn, something is bad with prow

PR #46, opened 2020-12-23
feat(sample): create sensible samples

georgettica:
now you can 'make sample-install' and 'make sample-uninstall' to work with samples

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -63,6 +63,14 @@ deploy: manifests kustomize
 	cd config/manager && $(KUSTOMIZE) edit set image controller=${IMG}
 	$(KUSTOMIZE) build config/default | kubectl apply -f -
 
+# Install CRDs into a cluster
+sample-install: manifests kustomize
+	$(KUSTOMIZE) build config/samples | kubectl apply -f -
+#
+# Install CRDs into a cluster
+sample-uninstall: manifests kustomize
+	$(KUSTOMIZE) build config/samples | kubectl delete -f -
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -63,6 +63,14 @@ deploy: manifests kustomize
 	cd config/manager && $(KUSTOMIZE) edit set image controller=${IMG}
 	$(KUSTOMIZE) build config/default | kubectl apply -f -
 
+# Install CRDs into a cluster
+sample-install: manifests kustomize
+	$(KUSTOMIZE) build config/samples | kubectl apply -f -
+#
+# Install CRDs into a cluster
```
RiRa12621, line 70: Uninstall?

RiRa12621 (changes requested):

Makefile
georgettica: Oh, copy pasta error.

Will.fix

config/samples/kustomization.yaml
```diff
@@ -1,3 +1,4 @@
 ## This file is auto-generated, do not modify ##
 resources:
+- monitoring_v1alpha1_clusterurlmonitor.yaml
```
RiRa12621, line 3: Why was that actually not present before?

georgettica: Nope, Manuel added it and it was not in the kustomization file (I guess he assumed it takes everything in the folder)

PR #51, opened 2021-01-17
fix(template): make deployment be in a different namespace

georgettica:
- 'make generate-syncset' now generates the namespace and the operatorgroup
- config/olm has a namespace parameter for overriding the namespace in the deployment pipeline

### DIFF
--- a/hack/olm-registry/olm-artifacts-template.yaml
+++ b/hack/olm-registry/olm-artifacts-template.yaml
@@ -6,7 +6,7 @@ parameters:
 - name: IMAGE_TAG
   required: true
 - name: NAMESPACE
-  value: openshift-monitoring
+  value: openshift-route-monitor-operator
 - name: REGISTRY_IMG
   required: true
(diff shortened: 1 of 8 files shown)


NautiluX (commented):
Do we need this to be configurable?
Also, will the blackbox-exporter deployment as well be created in the new namespace? Should it?

georgettica:
it is not, and we don't have to change that logic now IIUC, we just need that the operator will have full permissions to do what it needs to do

hack/olm-registry/olm-artifacts-template.yaml
```diff
@@ -6,7 +6,7 @@ parameters:
 - name: IMAGE_TAG
   required: true
 - name: NAMESPACE
-  value: openshift-monitoring
+  value: openshift-route-monitor-operator
```
georgettica, line 9: @NautiluX this is where we configure the namespace, it's just easier to manage and we can move it then later using the pipelines

PR #53, opened 2021-01-18
OSD-5709: Create script to extract RouteMonitor target URL

NautiluX:


RiRa12621:
missing a ticket or BZ ref and also a bit of context

RiRa12621:
@NautiluX can you add like a 1-2 line comment at the top so it's somewhat obvious what the script does, please?

NautiluX:
done

PR #54, opened 2021-01-19
Moving serviceMonitor to the CRs namespace

sulmer-redhat:
OSD-6240
OSD-6241

### DIFF
--- a/pkg/util/templates/templates.go
+++ b/pkg/util/templates/templates.go
@@ -1,26 +1,16 @@
 package templates
 
 import (
-	"fmt"
-
 	"github.com/openshift/route-monitor-operator/pkg/consts/blackbox"
 	monitoringv1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 )
 
-// TemplateForServiceMonitorName return the generated name from the RouteMonitor.
-// The name is joined by the name and the namespace to create a unique ServiceMonitor for each RouteMonitor
-func TemplateForServiceMonitorName(namespace string, name string) types.NamespacedName {
-	serviceMonitorName := fmt.Sprintf("%s-%s", name, namespace)
-	return types.NamespacedName{Name: serviceMonitorName, Namespace: blackbox.BlackBoxNamespace}
-}
-
 // TemplateForServiceMonitorResource returns a ServiceMonitor
-func TemplateForServiceMonitorResource(url, name string) monitoringv1.ServiceMonitor {
+func TemplateForServiceMonitorResource(url string, namespacedName types.NamespacedName) monitoringv1.ServiceMonitor {
 
 	routeURL := url
-	serviceMonitorName := name
 
 	routeMonitorLabels := blackbox.GenerateBlackBoxLables()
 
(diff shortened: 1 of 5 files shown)


RiRa12621:
/retitle Jira OSD-6240: Moving serviceMonitor to the CRs namespace

georgettica:
might be to another PR, but should we warn the user if his namespace doesn't have the correct label and won't be scraped?

that would make it much easier

pkg/util/templates/templates.go
```diff
@@ -1,26 +1,16 @@
...
-// TemplateForServiceMonitorName return the generated name from the RouteMonitor.
-// The name is joined by the name and the namespace to create a unique ServiceMonitor for each RouteMonitor
-func TemplateForServiceMonitorName(namespace string, name string) types.NamespacedName {
-	serviceMonitorName := fmt.Sprintf("%s-%s", name, namespace)
-	return types.NamespacedName{Name: serviceMonitorName, Namespace: blackbox.BlackBoxNamespace}
-}
-
 // TemplateForServiceMonitorResource returns a ServiceMonitor
-func TemplateForServiceMonitorResource(url, name string) monitoringv1.ServiceMonitor {
+func TemplateForServiceMonitorResource(url string, namespacedName types.NamespacedName) monitoringv1.ServiceMonitor {
```
georgettica, line 11: why not use the type https://github.com/openshift/route-monitor-operator/blob/d57e26c1f813ed740eb6b4a1281392d2fc942b75/api/v1alpha1/common_types.go#L4

it would make the code much cleaner

int/int_test.go
```diff
@@ -112,7 +111,7 @@ var _ = Describe("Integrationtests", func() {
 			err := i.RemoveRouteMonitor(routeMonitorNamespace, routeMonitorName)
 			routeMonitorName = "fake-route-monitor"
 			routeMonitorNamespace = "default"
-			expectedServiceMonitorName = templates.TemplateForServiceMonitorName(routeMonitorNamespace, routeMonitorName)
+			expectedServiceMonitorName = types.NamespacedName{Name: routeMonitorName, Namespace: routeMonitorNamespace}
```
georgettica, line 114: and you would just need to change from where we pull the resource
IIRC
```suggestion
			expectedServiceMonitorName = v1alpha1.NamespacedName{Name: routeMonitorName, Namespace: routeMonitorNamespace}
```

georgettica:
also it might be a good place to chage https://github.com/openshift/route-monitor-operator/blob/d57e26c1f813ed740eb6b4a1281392d2fc942b75/api/v1alpha1/routemonitor_types.go#L33
to be with the NamespacedName we have

pkg/util/templates/templates.go
NautiluX: sounds like a good idea!

int/int_test.go
georgettica: nvm, resending my comment

pkg/util/templates/templates.go
georgettica: oki, will do in another PR then :)

NautiluX: as discussed, since we need it in the `types.NamespaceName` type more often it's not worth the effort.

NautiluX:
@georgettica I don't think it's in the responsibility of this operator to inform the user they have to set up their prometheus correctly to have the ServiceMonitor scraped.

georgettica:
Ok.
Let's merge this one and continue work on a different PR

PR #67, opened 2021-01-25
Update issue templates

RiRa12621:


.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,34 @@
...
+
+**Expected behavior**
+A clear and concise description of what you expected to happen.
+
+**Screenshots**
+If applicable, add screenshots to help explain your problem.
+
+**OpenShift (please complete the following information):**
+ - Distribution: [e.g OCP]
+ - Client Version [e.g. `Client Version: 4.6.8`]
```
georgettica, line 28: Operator csv version?
Commit hash?

georgettica:
Where are these taken from?

RiRa12621:
GitHub defaults
Thought they're a good start

georgettica:
/approve
Let's see what someone else thinks

.github/ISSUE_TEMPLATE/bug_report.md
NautiluX: agree the client version is less interesting than the operator version. On the other hand that doesn't make sense until we have releases.

RiRa12621: that's actually why i let it out @NautiluX 
happy to adjust it as soon as we have a release

Tessg22: Is any user/customer running route monitor operator in non-test environment already?
If not, before deploying rmo on first production cluster by then we should have first release.
I agree with adding the operator version field. Eventually it will be useful.

georgettica: I think we currently have the csv version in the package manifest, so we can use it pre-production

PR #68, opened 2021-01-25
Create dependabot.yml

RiRa12621:


georgettica:
I thought we needed the dpp team to do that for us..
If this is the standard way I am up for it

RiRa12621:
They needed to enable it and I think they did so now we just configure it

coverprice:
Dependabot is already enabled for all `openshift` repos. However the default is to send notifications to the repo's admins, which in our case is not that useful because most teams are not admins on their own repos.

This `.github/dependabot.yaml` file can be configured by anyone with write access to the repo, so there is some level of self-service here.... however it's *not* possible to use this file to modify the Dependabot notification recipients; that's something that admins like DPP have to do through the GUI, for each repo.

coverprice:
Also, I'm not sure what you're trying to achieve with this particular `dependabot.yaml` config, but if it's just to "enable" it, that's not necessary. Dependabot is usually smart enough to work out the language + package deps files itself. Typically the only config needed is to update the notification recipients so that team members get the dependabot warnings/PRs, and DPP has to do that config.

RiRa12621:
Thanks for the answer @coverprice 
This is to configure dependabot to check the go dependencies daily, and submit a PR to update censored dependencies if needed

drpaneas:
+1. It's a great way to manage your Go dependencies in an automatic manner. You can also use it to update other GitHub actions (including itself). For example:

```yaml
  # Maintain dependencies for GitHub Actions
  - package-ecosystem: "github-actions"
    directory: "/"
    schedule:
      interval: "weekly"
    commit-message:
      prefix: "ci:"
```

NautiluX (approved):
/lgtm

PR #70, opened 2021-01-27
init(config): add prom-k8s config [OSD-6310]

NautiluX:


config/prometheus/monitor.yaml
```diff
@@ -1,4 +1,3 @@
-
```
RiRa12621, line 1: Where does that actually come from?

georgettica: idk.. magic?

PR #71, opened 2021-01-27
Fix csv

NautiluX:


hack/app_sre_create_image_catalog.sh
```diff
@@ -18,7 +18,7 @@ BUNDLE_DIR="$SAAS_OPERATOR_DIR/route-monitor-operator/"
 
 rm -rf "$SAAS_OPERATOR_DIR"
 
-git clone \
```
RiRa12621, line 21: Why>?

hack/bundler.sh
```diff
@@ -2,4 +2,4 @@
 set -euxo pipefail
 cd route-monitor-operator
 make packagemanifests
-chmod 775 -R packagemanifests
+# chmod 775 -R $BUNDLE_DIR
```
RiRa12621, line 5: That's going to bork the jenkins slave again

NautiluX: yes but I'm not yet sure how to get it right...

NautiluX: because it will change permissions of existing files. Not sure if that would even be a problem

NautiluX:
/hold
I want to observe this landing in the CI pipeline

config/manager/kustomization.yaml
```diff
@@ -4,4 +4,5 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 images:
 - name: controller
-  newName: route-monitor-operator
+  newName: controller
```
georgettica, line 7: Reset this please

hack/bundler.sh
```diff
@@ -2,4 +2,5 @@
 set -euxo pipefail
 cd route-monitor-operator
 make packagemanifests
-chmod 775 -R packagemanifests
+NEW_VERSION=$(ls "$BUNDLE_DIR" | sort -t . -k 3 -g | tail -n 1)
```
georgettica, line 5: This is good solution, but a better one would be taking the prev version from the yaml file in the saas.../rmo/*.yaml

NautiluX: this is not a good solution, it's a hack as the whole script.

NautiluX: The other way would be
```
grep currentCSV route-monitor-operator.package.yaml | sed 's/.*v'//g
```
but it doesn't look much better. Since my original version is the way we do it in the rest of the script, I'd prefer to leave it as is.

config/manager/kustomization.yaml
NautiluX: done

PR #74, opened 2021-01-28
Delete spec field from OperatorGroup

cblecker:
Got some non-matching operatorgroup errors on stage. This fixes them, and configures the operatorgroup for all namespaces (per the CSV spec).


RiRa12621:
/lgtm
/approve
Thanks @cblecker

NautiluX:
This change needs to be done [here](https://github.com/openshift/route-monitor-operator/blob/master/config/olm/operatorgroup.yaml) to survive `make generate-syncset`

PR #77, opened 2021-02-01
change the blackbox deployment namespace as per OSD-6298

sulmer-redhat:


georgettica:
/hold
can you look at something in the code?
we think that if someone changes the status.servicemonitorref it will cause a servicemonitor leak

RiRa12621:
/hold
@sulmer-redhat that doesn't look like a test flake but rather the change causing failing tests

georgettica:
@sulmer-redhat as this is related to #78 maybe close this?

PR #80, opened 2021-02-07
feat(blackbox): make blackbox prefer infra nodes [OSD-6337]

georgettica:
done via taints, tolerations and affinity


georgettica:
also, I verified and there is no const online for this infra node taint :/

NautiluX (commented):
/lgtm

PR #82, opened 2021-02-11
Jira OSD-5651: Boilerplate bootstrap

NautiluX:
* Changes necessary to onboard boilerplate in this repo
* Boilerplate bootstrap


.codecov.yml
```diff
@@ -0,0 +1,26 @@
+codecov:
```
RiRa12621, line 1: Do we really want this? I'm not the biggest fan of it due to the lack of accuracy

boilerplate/openshift/golang-osd-operator/.codecov.yml
```diff
@@ -0,0 +1,26 @@
+codecov:
```
RiRa12621, line 1: Do we really want this? I'm not the biggest fan of it due to the lack of accuracy

RiRa12621:
If this has any impact for users, it needs an addition to the readme

boilerplate/openshift/golang-osd-operator/.codecov.yml
2uasimojo: You mean because e.g. removing a line of code makes it look like you're decreasing coverage (e.g. 94/99 < 95/100)? Or because it's not taking generated code into account? Or something else? Is it the go cover tool specifically that you have issues with, or any coverage analysis?

My take is that doing it with inaccuracy is better than not doing it, since the limitations are known and we're not relying on the result at any serious level of precision. Currently we're [setting the bar ridiculously low](https://github.com/openshift/boilerplate/blob/459c3cdf9d1bf451f99a08fbc21833c817d540e1/boilerplate/openshift/golang-osd-operator/.codecov.yml#L8); if the inherent inaccuracies are making the difference between pass and fail, IMO you should be greatly increasing test coverage anyway.

Makefile
```diff
@@ -6,6 +8,14 @@ CURRENT_COMMIT=$(shell git rev-parse --short=7 HEAD)
...
+# for boilerplate
+OPERATOR_NAME=route-monitor-operator
+MAINPACKAGE=.
+TESTTARGETS=$(shell ${GOENV} go list -e ./... | egrep -v "/(vendor)/" | grep -v /int)
+
+# need to override boilerplate targets which are not working on this operator
+op-generate openapi-generate: ;
+
 VERSION ?= $(OPERATOR_VERSION)
 PREV_VERSION ?= $(VERSION)
```
2uasimojo, line 20: L33 (which I can't comment on directly): `IMG` is [set in standard.mk](https://github.com/openshift/route-monitor-operator/blob/97404eca7e5b669d904607c6a357d459fecb9116/boilerplate/openshift/golang-osd-operator/standard.mk#L28) as well. Since L33 comes after the `include`, but uses `?=`, it will be a no-op.

AFAICS this will break the [`pre-deploy`](https://github.com/openshift/route-monitor-operator/pull/82/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R74-R75) and [`podman-*`](https://github.com/openshift/route-monitor-operator/pull/82/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R105-R111) targets.

Makefile
```diff
@@ -91,13 +102,6 @@ vet:
...
-go-generate: mockgen
-	go generate ./...
-
-# Build the docker image
-docker-build: test
-	docker build . -t ${IMG}
-
 # Build the image with podman
 podman-build:
 	podman build . -t ${IMG}
```
2uasimojo, line 107: The [`docker-*` targets](https://github.com/openshift/route-monitor-operator/blob/97404eca7e5b669d904607c6a357d459fecb9116/boilerplate/openshift/golang-osd-operator/standard.mk#L87-L95) already use a [discovered, but overrideable `CONTAINER_ENGINE` variable](https://github.com/openshift/route-monitor-operator/blob/97404eca7e5b669d904607c6a357d459fecb9116/boilerplate/openshift/golang-osd-operator/standard.mk#L18-L19), so these (`podman-build` and `podman-push`) could be removed.

Makefile
```diff
@@ -91,13 +102,6 @@ vet:
 generate: controller-gen
 	$(CONTROLLER_GEN) object:headerFile="hack/boilerplate.go.txt" paths="./..."
```
2uasimojo, line 103: I'm not 100% sure, but I think the spirit of what this is doing is similar to what [boilerplate's `op-generate`](https://github.com/openshift/route-monitor-operator/blob/97404eca7e5b669d904607c6a357d459fecb9116/boilerplate/openshift/golang-osd-operator/standard.mk#L117-L124) does -- namely [`operator-sdk generate crds/k8s`](https://github.com/openshift/route-monitor-operator/blob/97404eca7e5b669d904607c6a357d459fecb9116/boilerplate/openshift/golang-osd-operator/operator-sdk-generate.sh#L46-L47) (kubebuilder etc).

As written, this overrides [boilerplate's `generate`](https://github.com/openshift/route-monitor-operator/blob/97404eca7e5b669d904607c6a357d459fecb9116/boilerplate/openshift/golang-osd-operator/standard.mk#L137-L138), which gets used in CI ([`validate`](https://github.com/openshift/route-monitor-operator/blob/97404eca7e5b669d904607c6a357d459fecb9116/boilerplate/openshift/golang-osd-operator/standard.mk#L182-L185) => [`generate-check`](https://github.com/openshift/route-monitor-operator/blob/97404eca7e5b669d904607c6a357d459fecb9116/boilerplate/openshift/golang-osd-operator/standard.mk#L155-L160)). If you could rename yours `op-generate` (and remove `op-generate` from [this override](https://github.com/openshift/route-monitor-operator/pull/82/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R17)), that just might plug it into the intended part of the chain.

Makefile
```diff
@@ -91,13 +102,6 @@ vet:
...
```
2uasimojo, line 104: [L134-47](https://github.com/openshift/route-monitor-operator/pull/82/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R134-R147) (which I can't comment on directly): You should be able to remove the `mockgen` target, as we [provide it](https://github.com/openshift/boilerplate/blob/459c3cdf9d1bf451f99a08fbc21833c817d540e1/config/build.sh#L59-L63) in the boilerplate backing image.

(We would be happy to update that image to include [`controller-gen`](https://github.com/openshift/route-monitor-operator/pull/82/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R119) and [`kustomize`](https://github.com/openshift/route-monitor-operator/pull/82/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R149) as well.)

Makefile
NautiluX: this is generated by  operator-sdk. I don't think `operator-sdk generate crds/k8s` is a valid command in operator-sdk v1

boilerplate/openshift/golang-osd-operator/.codecov.yml
NautiluX: are you saying we fail PRs if the lower the test coverage?

Makefile
NautiluX: the image doesn't seem to be used in the pipeline currently. Would be nice if `controller-gen`, `kustomize` and `operator-sdk` v1 were included there as well, probably a env variable could decide which version of the SDK to use

Makefile
NautiluX: done

Makefile
NautiluX: `pre-deploy` I guess will be called in most cases with a externally-set IMG variable so we should be fine removing the line.

Makefile
```diff
@@ -6,6 +8,14 @@ CURRENT_COMMIT=$(shell git rev-parse --short=7 HEAD)
 OPERATOR_VERSION=$(VERSION_MAJOR).$(VERSION_MINOR).$(COMMIT_NUMBER)-$(CURRENT_COMMIT)
```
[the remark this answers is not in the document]
NautiluX: done

Makefile
NautiluX: Guess we'll switch to that image as soon as it's supported, but it's nothing that we should worry about now.

boilerplate/openshift/golang-osd-operator/.codecov.yml
RiRa12621: What I mean is, that I don't see a benefit in having it. Especially given the current coverage

Makefile
2uasimojo: Right, it's not, my point is that we want `make op-generate` to do whatever the v1 equivalent is; and we want `make generate` to be the broader target that includes go and openapi (and any other generators we add in the future).

Makefile
2uasimojo: > the image doesn't seem to be used in the pipeline currently.

Ah, right, that'll come in when we standardize prow config. See for example [this recent PR](https://github.com/openshift/release/pull/15470) for managed-upgrade-operator. It's [this bit](https://github.com/openshift/release/pull/15470/files#diff-06907a927697b3256104433ace45af68ffecb69ba5e61f421c692166ab151758R2) that makes prow pay attention to the [.ci-operator config](https://github.com/openshift/route-monitor-operator/pull/82/files#diff-4d5699538541ad6da5f6ae79150b49a432b610f4cba488cec683ac9dd00b18d2) in this PR, which is what grabs that image.

> Would be nice if `controller-gen`, `kustomize` and `operator-sdk` v1 were included there as well

Agreed. That would all go in [this script](https://github.com/openshift/boilerplate/blob/master/config/build.sh), which has several lovely examples to follow.

Specifically for osdk, we install all the supported versions in the image (via that script). The make targets run them via [this script](https://github.com/openshift/boilerplate/blob/459c3cdf9d1bf451f99a08fbc21833c817d540e1/boilerplate/openshift/golang-osd-operator/operator-sdk-generate.sh), which would need to be modified to include a condition for v1 with the appropriate syntax. Also see below:

> probably a env variable could decide which version of the SDK to use

The way it works today, we [discover your desired osdk version via go.mod](https://github.com/openshift/boilerplate/blob/459c3cdf9d1bf451f99a08fbc21833c817d540e1/boilerplate/openshift/golang-osd-operator/ensure.sh#L41). IIUC with v1 there's not actually a go dependency on osdk itself, so we'll need to either a) add an `indirect` one; or b) figure out some other mechanism.

2uasimojo:
Thanks for the updates. I think once we resolve the [`make generate`](https://github.com/openshift/route-monitor-operator/pull/82#discussion_r574633565) issue, this will be ready to go. (There will be followup work once prow config lands, as discussed above.)

boilerplate/openshift/golang-osd-operator/.codecov.yml
2uasimojo: > are you saying we fail PRs if the lower the test coverage?

Yes. The cutoff is currently 20%. We hope to be able to raise that as time goes on. But it's understood that there's no expectation to test e.g. generated controller framework code, so it'll never be 100%. (As one example, I feel like [aws-efs-operator](https://codecov.io/gh/openshift/aws-efs-operator) has pretty much all of its non-generated code covered, and it shows just north of 75%.)

> What I mean is, that I don't see a benefit in having it. Especially given the current coverage

Are you saying the current coverage is high or low? (I didn't check.) Or are you saying that the integration tests, which wouldn't run under coverage, obviate the need for unit tests?

In general the benefits include:
- It gives reviewers a hint as to whether you forgot to include unit tests with your PR (if coverage goes down significantly).
- It provides nice visual dashboards that can help you easily find hot spots of covered/uncovered code.

If you want to ignore the reports as a team, that's up to you. If you seriously have <20% unit test coverage currently, you may want to consider remedying that -- integration tests notwithstanding. But the whole point of boilerplate is to avoid snowflaking individual consumers. So I'd really like to keep this in place.

2uasimojo:
> If this has any impact for users, it needs an addition to the readme

What do you mean by "users"? There should be no end user impact. For developers/deployers, I believe the [existing README content](https://github.com/openshift/route-monitor-operator/#development) is still accurate. More could be added there in general, but that'd be for another PR.

build/Dockerfile
```diff
@@ -0,0 +1,30 @@
+FROM quay.io/app-sre/boilerplate:image-v0.4.0 AS builder
+FROM registry.svc.ci.openshift.org/openshift/release:golang-1.13 AS builder
```
2uasimojo, line 2: Just noticed this. We can get rid of this second `FROM` line. (The boilerplate image is based on golang-1.13)

Makefile
NautiluX: I think all generation commands are provided by the SDK, no reason to use one from boilerplate.

build/Dockerfile
NautiluX: done

boilerplate/openshift/golang-osd-operator/.codecov.yml
NautiluX: A CI should never fail based on code coverage, probably we should fix that in boilerplate. What's the reason to block a PR that lowers the coverage from 20% to 19% while one that lowered it from 30% to 29% is fine? Probably it's a theoretical issue, since we don't have that low coverage in this repository, but I think failing CI based on code coverage is an antipattern. I'm fine with making it visible in the PR conversation, though.

.codecov.yml
NautiluX: this discussion needs to be brought to boilerplate.

boilerplate/openshift/golang-osd-operator/.codecov.yml
jharrington22: The 20% threshold is there to encourage the team to add tests, its not a high bar and we could also change it. I would agree with you @NautiluX if the threshold were > 70% and CI tests were failing all the time. I don't see the hard here. 

@RiRa12621 If you have a specific issue with codecov let us know, sounds like you had some questions around the accuracy but I'm also not sure what you mean.

jharrington22:
Hey @NautiluX what is the status here with this PR? We are working towards onbaording all the operator operators to use the OLM builds within boilerplate, this is a requirement to support image digests so we need to do the same here. How can we help?

NautiluX:
> Hey @NautiluX what is the status here with this PR? We are working towards onbaording all the operator operators to use the OLM builds within boilerplate, this is a requirement to support image digests so we need to do the same here. How can we help?

Hi @jharrington22 this was waiting on an updated image I was working on together with @2uasimojo. This is done now so I'll go rebase/update this.

NautiluX:
actually the image update can wait. Let's push this through first and update the image in a separate PR

jharrington22:
> > Hey @NautiluX what is the status here with this PR? We are working towards onbaording all the operator operators to use the OLM builds within boilerplate, this is a requirement to support image digests so we need to do the same here. How can we help?
> 
> Hi @jharrington22 this was waiting on an updated image I was working on together with @2uasimojo. This is done now so I'll go rebase/update this.

No worries, is there a reason you can't update the image in this PR?

NautiluX:
oh, I updated the image, the idea is that we can replace the custom image we currently use in the pipeline with the boilerplate one which will need some more testing. So I thought it makes sense to get this done independently. (We also track it in 2 independent stories in our backlog)

2uasimojo:
That test failure happened twice in a row. I don't know exactly what it's doing, but it seems to be running afoul of an ellipsis unicode character (`…`) in a log somewhere. That doesn't smell like a flake, does it?

jharrington22:
@NautiluX @RiRa12621 If you could take a look first thing tomorrow that would be appreciated.

NautiluX:
The tests succeed, it's a infrastructure problem, had the same on another PR.
```
Step integration-aws-test succeeded after 3m50.094048205s
```

boilerplate/openshift/golang-osd-operator/.codecov.yml
RiRa12621: I bet now you're happy about this comment 😜

2uasimojo:
This could land if you're satisfied with it, but we still have serious issues with the boilerplate-isms. At a minimum:
- The appsre pipeline stuff *will not work*. As long as you don't change the [ci-int config](https://gitlab.cee.redhat.com/service/app-interface/-/blob/7ca26dc5cbe596cecb3131b120779f5adbde962b/data/services/osd-operators/cicd/ci-int/jobs-route-monitor-operator.yaml#L27) you'll continue using your bespoke osdk v1 csv gen, so that might be okay.
- There's still the conflict with generators as noted inline: osdk is greedily slurping up the `generate` target, which boilerplate uses to subsume generators for several things (osdk, go, openapi).
- That, and other things, will make it difficult to standardize the prow configuration.

TL;DR: with all the discrepancies I'm not sure we're getting enough benefit out of boilerplate to make it worth incorporating at all at this time.

We absolutely need to solve pulling the operator and registry images by digest, but if we try to do that with boilerplate I believe we'll miss our target. We'll need to work on doing that "manually".

NautiluX:
I'd say we first get this in and fix further things down the road separately? I think the generate of boilerplate does (mainly/only) things which are not needed for v1 operators so probably it's ok to override them for the time being.
CSV generation works different for v1, I think we need to find a way to detect it's v1 and run a different procedure.

2uasimojo:
Okay then

/lgtm

for boilerplate

/hold

so you can control when it lands.

PR #90, opened 2021-02-22
fix(servicemonitor): make servicemonitor work [OSD-6541]

georgettica:
still need to test on a cluster to see if it looks pretty


NautiluX (commented):
/lgtm
/hold
remove hold when manual testing is done

georgettica:
Did manual testing and it does remove the job and is scrapable.
Sidenote, i had two routemonitors and and for some reason only one had visible metrics in prometheus

georgettica:
/unhold
as I mentioned yesterday, tested (also using olm) and it works

PR #94, opened 2021-03-04
Allowing for SLO changes

sulmer-redhat:


controllers/routemonitor/routemonitor_test.go
```diff
@@ -42,18 +42,20 @@ var _ = Describe("Routemonitor", func() {
...
-		get                                   helper.MockHelper
-		create                                helper.MockHelper
-		ensureServiceMonitorResourceAbsent    helper.MockHelper
-		ensurePrometheusRuleResourceAbsent    helper.MockHelper
-		shouldDeleteBlackBoxExporterResources helper.MockHelper
-		ensureBlackBoxExporterResourcesAbsent helper.MockHelper
-		ensureBlackBoxExporterResourcesExist  helper.MockHelper
-		ensureFinalizerAbsent                 helper.MockHelper
-
+		mockResourceComparer         *routemonitormocks.MockResourceComparer
```
georgettica, line 45: NIT: move to where all the mocks are (above ctx)

controllers/routemonitor/routemonitor_test.go
```diff
@@ -42,18 +42,20 @@ var _ = Describe("Routemonitor", func() {
...
+		delete                                        helper.MockHelper
+		get                                           helper.MockHelper
+		create                                        helper.MockHelper
+		ensureServiceMonitorResourceAbsent            helper.MockHelper
+		ensurePrometheusRuleResourceAbsent            helper.MockHelper
+		shouldDeleteBlackBoxExporterResources         helper.MockHelper
+		ensureBlackBoxExporterResourcesAbsent         helper.MockHelper
+		ensureBlackBoxExporterResourcesExist          helper.MockHelper
+		ensureFinalizerAbsent                         helper.MockHelper
+		deepEqual                                     helper.MockHelper
```
georgettica, line 57: not sure we actually want to keep this as helper.MockHelper as it's only used for the int inside it

maybe change to (and fix corresponding places)
```suggestion
		deepEqualCalledTimes                                     int
```

config/manager/kustomization.yaml
```diff
@@ -4,5 +4,5 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 images:
 - name: controller
-  newName: route-monitor-operator
+  newName: controller
```
georgettica, line 7: this has hit us multiple times.. maybe we should have some validation on this file / where we change this field?

https://github.com/openshift/route-monitor-operator/blob/229da76cc133b1226eb69ecb1a7aaafa70563c92/Makefile#L64

georgettica (commented):
a very tiny nit and overall we worked together on it so I +1 this

should I lgtm aswell?

georgettica:
sorry about that @sulmer-redhat (merge conflict) , if you want we can sit on this later next week to resolve

sulmer-redhat:
Closing in favour of !99.

PR #95, opened 2021-03-04
feat(routemonitor): add errorStatus [OSD-6306]

georgettica:
- this makes a user misocnfiguration show up in '.status.errorStatus'


georgettica:
🤞 it works

RiRa12621:
/lgtm
if it works, it's good

PR #97, opened 2021-03-08
add burn rate to alerts

NautiluX:
* Add burn rate alerts to longer windows (compare with https://sre.google/workbook/alerting-on-slos/#recommended_parameters_for_an_slo_based_a)
* Use avg_over_time for a proper alert threshold


RiRa12621:
Average over time is flattening thus a lot.
Not sure that is the way.

PR #98, opened 2021-03-18
fix(sample): change SLA to correct value

georgettica:
change according to reqs


RiRa12621 (approved):
/lgtm

NautiluX (commented):
we'll need to revert this PR if we care for our samples.

PR #99, opened 2021-03-22
georgettica/more changes

georgettica:
Allowing for SLO changes

* Allowing SLOs to be changed within the routeMonitor CR
* Adding unit tests, integration tests

### DIFF
--- a/int/integration.go
+++ b/int/integration.go
@@ -4,7 +4,8 @@ import (
 	"context"
 	"fmt"
 	"time"
-
+	"reflect"
+	"github.com/openshift/route-monitor-operator/pkg/util/templates"
 	"github.com/onsi/ginkgo"
 	configv1 "github.com/openshift/api/config/v1"
 	routev1 "github.com/openshift/api/route/v1"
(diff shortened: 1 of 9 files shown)


georgettica:
/hold
in case this passes too fast

georgettica:
This is holding for additional test& integration..
If we want to expose the way we do it maybe additing a local SOP for this is a good solution and suggest it to a base location later on

sulmer-redhat:
Added Integration tests!

int/integration.go
```diff
@@ -186,3 +187,41 @@ func (i *Integration) WaitForPrometheusRuleToClear(name types.NamespacedName, se
...
+	routeMonitor := v1alpha1.RouteMonitor{}
+	err = i.Client.Get(context.TODO(), name, &routeMonitor)
+	if errors.IsNotFound(err) {
+		return fmt.Errorf("RouteMonitor wasn't found")
+	}
+	if err != nil {
+		return err
+	}
+
+	template := templates.TemplateForPrometheusRuleResource(routeMonitor.Status.RouteURL, targetSlo, name)
```
georgettica, line 210: Why should we create the template of we have the routemonitors before any action was done?

georgettica:
It looks ok, and I don't want to keep it on for longer so I think we can approve this..

NautiluX:
Is there any story this is related to? the commit messages don't explain too much.

georgettica:
It's a fork from #94, and the odd can be added by @sulmer-redhat

int/integration.go
sulmer-redhat: Because inside the loop we're going to compare the actual PrometheusRule to the template, to check and see if it is being updated. I didn't want to generate the template from within the loop to avoid calling the "TemplateForPrometheusRuleResource" multiple times for no reason.

georgettica:
/lgtm
let's merge and keep moving forward!

georgettica:
@sulmer-redhat if you want, you can lgtm the PR as you are not the "owner" of it

georgettica:
/unhold
let it rip!

georgettica:
@sulmer-redhat maybe the test is flakey?

NautiluX:
the tests succeeded, seems to be an issue with test infra, it fails after the tests:
```
'ascii' codec can't encode character '\u2026' in position 6438: ordinal not in range(128) 
INFO[2021-04-14T10:58:05Z] {"component":"entrypoint","error":"wrapped process failed: exit status 1","file":"prow/entrypoint/run.go:80","func":"k8s.io/test-infra/prow/entrypoint.Options.Run","level":"error","msg":"Error executing test process","severity":"error","time":"2021-04-14T10:58:04Z"} 
```

PR #100, opened 2021-03-26
WIP: Tekton Pipeline Objects

NautiluX:


georgettica:
does this work better than `test-integration`?

NautiluX (commented):
Thanks for this PR, can you add a short readme describing how to use it?

NautiluX:
> does this work better than `test-integration`?

this will run on a cluster that has RMO deployed already, it's a different usecase than the test-integration script. You still have issues with the GCP deployed CRC?

georgettica:
The only issue is that passing the image up to gxp takes forever so I use https://www.cis.upenn.edu/~bcpierce/unison/ to copy to gxp and run `test-integration` there

PR #102, opened 2021-04-05
Pipelines

ramonbutter:
Resources for running an automated tekton pipeline that tests the RMO installation


RiRa12621:
Can someone enlighten me why we are adding Tekton stuff?

ramonbutter:
@RiRa12621 Its described in ticket OSD-6823

RiRa12621:
It is actually not. There's a tekton example PR but that's about it.
I don't see a justification in there to introduce yet an other CI solution next to Jenkins and Prow

NautiluX:
It doesn't introduce anything, it provides a bunch of Pipelines you can use to simply run our tests on any OCP cluster.

RiRa12621:
/approve
merge at will

ci/README.md
```diff
@@ -0,0 +1,48 @@
+# Tekton CI Pipeline
+
+This directory contains the configuration for a tekton pipeline that tests
+the installed `route-monitor-operator` version in the logged in cluster.
+
+## Installation
+
+A `route-monitor-operator` installation has to be present.
+The rest of the installation is done by running the following command several times:
```
NautiluX, line 9: Why does the command have to be run several times?

ci/README.md
```diff
@@ -0,0 +1,48 @@
+# Tekton CI Pipeline
+
+This directory contains the configuration for a tekton pipeline that tests
+the installed `route-monitor-operator` version in the logged in cluster.
```
NautiluX, line 4: are there other things that need to be installed on the cluster before we can run it? I.e. is Tekton part of it?

NautiluX: Is there is a yaml for the subscription, applying this should probably be a first, independent step.

NautiluX (changes requested):

ci/README.md
ramonbutter: Because some of the yamls depend on each other. First, the pipeline operator needs to be installed, then the service account. Then the tekton CRs can be created etc... Sure I could make an own script for that, but it shouldn't matter to much. As running `oc apply -f .` is idempotent

ci/README.md
ramonbutter: subscription to tekton is also defined in [this yaml](https://github.com/openshift/route-monitor-operator/pull/102/files#diff-ffa7e8f720f478144b6ed0d308aa83597c25514617a0fa394a8e3d79e5e497e7). And I would say the same argument as above applies

NautiluX: yeah I think I'd prefer just mention applying that subscription first and then the pipeline stuff rather than apply until good ;)

NautiluX (commented):
/lgtm

PR #103, opened 2021-04-06
Pin yq to 3.4.1

2uasimojo:
See openshift/osd-metrics-exporter#43 for details.


RiRa12621:
Shouldn't we rather work towards 4.x usability as well instead of pinning an old version?

2uasimojo:
> Shouldn't we rather work towards 4.x usability as well instead of pinning an old version?

What we should do is finish getting this [functionality into boilerplate](https://github.com/openshift/boilerplate/pull/119) and then have consumers adopt it, so when we want to do something like "work towards 4.x" we only have to do it in one place.

For now, this will get RMO deploying again.

PR #105, opened 2021-04-19
Pull operator image by digest

NautiluX:


hack/app_sre_build_deploy.sh
```diff
@@ -10,24 +10,38 @@ BASE_IMG="route-monitor-operator"
 QUAY_IMAGE="${QUAY_IMAGE:-quay.io/app-sre/${BASE_IMG}}"
 IMG="${BASE_IMG}:latest"
```
2uasimojo, line 11: This line can go away now

hack/app_sre_build_deploy.sh
```diff
@@ -10,24 +10,38 @@ BASE_IMG="route-monitor-operator"
...
-  skopeo copy --dest-creds "${QUAY_USER}:${QUAY_TOKEN}" \
-      "docker-daemon:${IMG}" \
-      "docker://${QUAY_IMAGE}:latest"
-
-  skopeo copy --dest-creds "${QUAY_USER}:${QUAY_TOKEN}" \
-      "docker-daemon:${IMG}" \
-      "docker://${QUAY_IMAGE}:${GIT_HASH}"
+OPERATOR_IMAGE_URI=${QUAY_IMAGE}:${GIT_HASH}
+BUILD_CMD="docker build" IMG="$OPERATOR_IMAGE_URI" make docker-build
+REGISTRY_IMAGE="quay.io/app-sre/route-monitor-operator-registry"
```
2uasimojo, line 22: Suggest `:-`ing this as well so you can override it from the env for testing.

2uasimojo (commented):
Yeah, this seems sane to me.

2uasimojo:
/lgtm
/hold

Unhold when ready. If you apply the second change suggested inline, you should be able to test this locally. LMK if you want help with that. [This doc](https://github.com/openshift/boilerplate/blob/master/boilerplate/openshift/golang-osd-operator/app-sre.md) will be a good start -- but you'll need the extra overrides for `QUAY_IMAGE` and `REGISTRY_IMAGE`.

2uasimojo:
With a couple of tweaks, I was able to test this locally and validate that it works as expected. The before/after delta in the bundles looks like:

```diff
diff -r route-monitor-operator/0.1.260-c12ed28/route-monitor-operator.clusterserviceversion.yaml route-monitor-operator/0.1.275-d08929e/route-monitor-operator.clusterserviceversion.yaml
9c9
<   name: route-monitor-operator.v0.1.260-c12ed28
---
>   name: route-monitor-operator.v0.1.275-d08929e
176c176
<                 image: quay.io/app-sre/route-monitor-operator:c12ed28
---
>                 image: controller@sha256:ec8a3eeb9684d3cb400ccc452982961d42c920a81cb0910d6b5dc614240d22be
251,252c251,252
<   replaces: route-monitor-operator.v0.1.256-82f32d8
<   version: 0.1.260-c12ed28
---
>   replaces: route-monitor-operator.v0.1.260-c12ed28
>   version: 0.1.275-d08929e
```

Which is exactly what we want.

I was also able to validate that the "did we already do this? If so, skip it." logic works properly.

/lgtm

Unhold when ready.

2uasimojo:
> ```diff
> image: controller@sha256:ec8a3eeb9684d3cb400ccc452982961d42c920a81cb0910d6b5dc614240d22be
> ```

Oops, this is bad, sorry I missed it. #106 for the fix.

PR #107, opened 2021-04-20
OSD-6853: CatalogSource by digest

2uasimojo:
[OSD-6853](https://issues.redhat.com/browse/OSD-6853)

### DIFF
--- a/hack/olm-registry/olm-artifacts-template.yaml
+++ b/hack/olm-registry/olm-artifacts-template.yaml
@@ -5,6 +5,8 @@ parameters:
   required: true
 - name: IMAGE_TAG
   required: true
+- name: REPO_DIGEST
+  required: true
 - name: NAMESPACE
   value: openshift-route-monitor-operator
 - name: REGISTRY_IMG
@@ -48,7 +50,7 @@ objects:
         icon:
           base64data: ''
           mediatype: ''
-        image: ${REGISTRY_IMG}:${CHANNEL}-${IMAGE_TAG}
+        image: ${REPO_DIGEST}
         publisher: Red Hat
         sourceType: grpc
     - apiVersion: operators.coreos.com/v1alpha1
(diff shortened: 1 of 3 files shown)


georgettica:
Hey, the REPO_DIGEST is not in the template file and thus won't work when replacing.

There is a process that deploys RMO using the template (so you can test this works)

Would you want me to outline it in a different PR so you can follow?
If not would you like me to test it to verify it's working as expected?

2uasimojo:
/hold

Whoah, one of my deltas didn't make it. Weird.

2uasimojo:
> make generate-syncset

[This automated commit](https://github.com/openshift/route-monitor-operator/pull/107/commits/fa50b8aefc44aa29e035ad1fcb1aa3ee979f7107) removed the required var from the template. WTF?

2uasimojo:
> Hey, the REPO_DIGEST is not in the template file and thus won't work when replacing.
> 
> There is a process that deploys RMO using the template (so you can test this works)
> 
> Would you want me to outline it in a different PR so you can follow?
> If not would you like me to test it to verify it's working as expected?

I think I figured it out @georgettica -- can you please confirm on the latest rev?

hack/olm-registry/olm-artifacts-template.yaml
```diff
@@ -5,6 +5,8 @@ parameters:
   required: true
 - name: IMAGE_TAG
   required: true
+- name: REPO_DIGEST
+  required: true
```
2uasimojo, line 9: ✓

2uasimojo (commented):
/hold cancel

This is better.

georgettica:
> > Hey, the REPO_DIGEST is not in the template file and thus won't work when replacing.
> > There is a process that deploys RMO using the template (so you can test this works)
> > Would you want me to outline it in a different PR so you can follow?
> > If not would you like me to test it to verify it's working as expected?
> 
> I think I figured it out @georgettica -- can you please confirm on the latest rev?

didn't see it

more active on other channels

PR #109, opened 2021-05-04
OSD-6979: changed PrometheusRule Template

ramonbutter:


pkg/util/templates/templates.go
```diff
@@ -86,9 +91,18 @@ type multiWindowMultiBurnAlertRule struct {
...
+	lWindow, _ := time.ParseDuration(r.longWindow)
+	mPeriod, _ := time.ParseDuration(serviceMonitorPeriod)
+
+	alertString := "1-(sum(sum_over_time(probe_success{" + label + "}[" + r.shortWindow + "])) / sum(count_over_time(probe_success{" + label + "}[" + r.shortWindow + "])))" + 
+	"> (" + r.burnRate + "*(1-" + percent + "))\n" + 
+	"and sum(count_over_time(probe_success{" + label + "}[" + r.shortWindow + "])) > " + fmt.Sprintf("%d", int(sWindow.Minutes()/mPeriod.Minutes()/2)) +
+	"\nand\n" +
+	"1-(sum(sum_over_time(probe_success{" + label + "}[" + r.longWindow + "])) / sum(count_over_time(probe_success{" + label + "}[" + r.longWindow + "])))" + 
+	"> (" + r.burnRate + "*(1-" + percent + "))\n" + 
+	"and sum(count_over_time(probe_success{" + label + "}[" + r.longWindow + "])) > " + fmt.Sprintf("%d", int(lWindow.Minutes()/mPeriod.Minutes()/2))
```
NautiluX, line 104: for future us: can you try to make this a bit more readable/understandable?

pkg/util/templates/templates.go
```diff
@@ -86,9 +91,18 @@ type multiWindowMultiBurnAlertRule struct {
 			 targeturl: getmeright
 */
 func (r *multiWindowMultiBurnAlertRule) render(url, percent, label, alertName string) monitoringv1.Rule {
-	alertString := "1-avg_over_time(probe_success{" + label + "}[" + r.shortWindow + "]) > (" + r.burnRate + "*(1-" + percent + "))\n" +
-		"and\n" +
-		"1-avg_over_time(probe_success{" + label + "}[" + r.longWindow + "]) > (" + r.burnRate + "*(1-" + percent + " ))"
+	sWindow, _ := time.ParseDuration(r.shortWindow)
+	lWindow, _ := time.ParseDuration(r.longWindow)
+	mPeriod, _ := time.ParseDuration(serviceMonitorPeriod)
```
NautiluX, line 96: is it safe to ignore the errors returned?

ramonbutter: I would say so. These values are only set by us, and an error doesn't need to be necessarily handled here as one would fail the unit or integration test and shouldn't be merged anyways

pkg/util/templates/templates.go
ramonbutter: Would you like me to refactor it?

NautiluX: yes please

pkg/util/templates/templates.go
```diff
@@ -70,6 +76,29 @@ type multiWindowMultiBurnAlertRule struct {
...
+
+	return rule
+}
+
+func sufficientProbes(windowSize, label string) string {
+	window, _ := prometheus.ParseDuration(windowSize)
+	window_duration := time.Duration(window)
+	mPeriod, _ := prometheus.ParseDuration(serviceMonitorPeriod)
+	mPeriod_duration := time.Duration(mPeriod) 
+	necessarySamplesInWindow := 0.5
```
NautiluX, line 94: this sounds like a number of samples you expect, but it's only the percentage of sample needed. What do you think of moving the calculation from the Sprintf to this line? I think the readability of promrule could benefit from that.

pkg/util/templates/templates.go
```diff
@@ -86,9 +115,16 @@ type multiWindowMultiBurnAlertRule struct {
...
-		"1-avg_over_time(probe_success{" + label + "}[" + r.longWindow + "]) > (" + r.burnRate + "*(1-" + percent + " ))"
+
+	alertString := "" +
+		alertThreshold(r.shortWindow, percent, label, r.burnRate) +
+		" and " +
+		sufficientProbes(r.shortWindow, label) +
+		"\nand\n" +
+		alertThreshold(r.longWindow, percent, label, r.burnRate) +
+		" and " +
+		sufficientProbes(r.longWindow, label)
```
NautiluX, line 126: much better

pkg/util/templates/templates.go
```diff
@@ -70,6 +76,29 @@ type multiWindowMultiBurnAlertRule struct {
...
+
+func sufficientProbes(windowSize, label string) string {
+	window, _ := prometheus.ParseDuration(windowSize)
+	window_duration := time.Duration(window)
+	mPeriod, _ := prometheus.ParseDuration(serviceMonitorPeriod)
+	mPeriod_duration := time.Duration(mPeriod) 
+	necessaryProbesInWindow := int(window_duration.Minutes()/mPeriod_duration.Minutes() * 0.5)
+
+	rule := "sum(count_over_time(probe_success{" + label + "}[" + windowSize + "]))" +
+		" > " + fmt.Sprintf("%d", necessaryProbesInWindow)
```
NautiluX, line 97:
```suggestion
		" > " + necessaryProbesInWindow
```

pkg/util/templates/templates.go
```diff
@@ -70,6 +76,29 @@ type multiWindowMultiBurnAlertRule struct {
...
+
+	rule := "sum(count_over_time(probe_success{" + label + "}[" + windowSize + "]))" +
+		" > " + strconv.Itoa(necessaryProbesInWindow)
+
+	return rule
+}
+
 //render creates a monitoring rule for the defined multiwindow multi-burn rate alert
 // Sample result as yaml
 /*
```
NautiluX, line 104: this comment seems super outdated now

ramonbutter: yes I would also just drop these kind of comments

PR #110, opened 2021-05-08
OSD-6726: Adding ClusterID to ServiceMonitorLabels

ramonbutter:


ramonbutter:
We are doing similar for other metrics. Its filling a "_id" or "clusterID" label. Somehow I dont see any issue.

RiRa12621:
clusterID can be ignored when we talk about the cardinality.

image you have a metrics `probe` and it used to have labels `result` and `superLabel` that can have different values. We can assume that the cardinality is defined as possible values for result * possible values for superLabel. If we add a ClusterID into the equation, that can have a maximum of 1 different values so you'd multiply by one. tl;dr clusterID doesn't impact the cardinality.

ramonbutter:
\hold

pkg/util/templates/templates.go
```diff
@@ -162,3 +169,13 @@ func sampleTemplateLabelsWithSev(url, severity string) map[string]string {
...
 	}
 }
+
+func GetClusterID(c client.Client) string {
+	var version configv1.ClusterVersion
+	fmt.Println("Called GetClusterID")
+	err := c.Get(context.TODO(), client.ObjectKey{Name: "version"}, &version)
+	if err != nil {
+		return ""
+	}
```
NautiluX, line 179: Shouldn't we try getting the ID again with the next reconcile instead of ignoring the error?

NautiluX (changes requested):

controllers/clusterurlmonitor/clusterurlmonitor_supplement.go
```diff
@@ -59,10 +65,13 @@ func (s *ClusterUrlMonitorSupplement) EnsureServiceMonitorExists() error {
 	if err != nil {
 		return err
 	}
+	if s.ClusterID == "" {
+		s.ClusterID = templates.GetClusterID(s.Client)
+	}
```
NautiluX, line 70: won't `s.ClusterID` be empty on every reconcile?

controllers/clusterurlmonitor/clusterurlmonitor_supplement.go
```diff
@@ -32,10 +32,16 @@ type ClusterUrlMonitorSupplement struct {
...
-	return &ClusterUrlMonitorSupplement{ClusterUrlMonitor, Client, log, context.Background(), blackBoxExporter}
+	return &ClusterUrlMonitorSupplement{
+		ClusterUrlMonitor: ClusterUrlMonitor,
+		Client:            Client,
+		Log:               log,
+		Ctx:               context.Background(),
+		BlackBoxExporter:  blackBoxExporter}
 }
 
 func (s *ClusterUrlMonitorSupplement) EnsureFinalizer() (reconcile.Result, error) {
```
NautiluX, line 47: is this function the right place for the functionality? How does the ClusterID label relate to the finalizer?

pkg/util/templates/templates.go
ramonbutter: For the case something fails, it is set to an empty string, and it will try to fetch it again in the next reconcile. However, there is actually not a big chance that this fails as long as API is up and Permissions are set

controllers/clusterurlmonitor/clusterurlmonitor_supplement.go
ramonbutter: Only after the first reconcile and in case an error happend during the GetClusterID. In this case it will try to fetch the ClusterID again

controllers/clusterurlmonitor/clusterurlmonitor_supplement.go
ramonbutter: Ok thats a good question, don't know much about the finalizer. I didn't change that part. But I would assume it doesn't influence the Finalizer list? As it ClusterID is a primitive data type?

controllers/clusterurlmonitor/clusterurlmonitor_supplement.go
ramonbutter: And Supplement is just allocated once [here](https://github.com/openshift/route-monitor-operator/blob/4abfff4b9a4ba5552c48d29db47e965c2e050b4c/main.go#L102), so values are maintained between reconcile loops

ramonbutter: ok no you are right, its not for the clusterurlmonitor_supplement but just for the routemonitor.
A new clusterurl Supplement is allocated on each Reconcile. Thats a bit unfortunate. However, I could move the ClusterID fetching and it would have the same effect. However, what do you think. Is this approach ok, or should we go with altering the BlackBoxExporter Deployment CR ?

controllers/clusterurlmonitor/clusterurlmonitor_supplement.go
NautiluX: It's more that I wouldn't want the function to set an unrelated variable as a side effect.

controllers/clusterurlmonitor/clusterurlmonitor_supplement.go
NautiluX: I don't mind too much at which object it is.

pkg/util/templates/templates.go
NautiluX: I _guess_ there could be more and uninteresting reasons the k8s api call could fail

georgettica:
@ramonbutter the blocker PR is no longer blocking

(#112) 
and feel free to work with what I have in #119

ramonbutter:
Yep thanks, I am having a merging mess, but hope to be able to finish this thing until the evening

int/int_test.go
```diff
@@ -408,7 +408,11 @@ var _ = Describe("Integrationtests", func() {
 				err = i.Client.Update(context.TODO(), &latestRouteMonitor)
 				Expect(err).NotTo(HaveOccurred())
 				_, parsedSlo := latestRouteMonitor.Spec.Slo.IsValid()
+<<<<<<< HEAD
 				err = i.RouteMonitorWaitForPrometheusRuleCorrectSLO(expectedDependentResource, parsedSlo, 20, routeMonitor.Kind)
+=======
+				err = i.WaitForPrometheusRuleCorrectSLO(expectedDependentResource, parsedSlo, 20)
+>>>>>>> 229319a (added ClusterID to templates)
```
NautiluX, line 415: I think we have an unresolved merge conflict here

NautiluX (changes requested):

int/int_test.go
ramonbutter: Yes, I just wanted to save my changes before I went on vacation. It's very much still in Progress. Unfortunately, I lost myself a bit in refactoring, which takes just an insane amount of time although I wanted to keep it simple.

controllers/routemonitor/routemonitor.go
```diff
@@ -68,23 +95,19 @@ func (r *RouteMonitorReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error
...
-	res, err = r.EnsureFinalizerSet(ctx, routeMonitor)
-	if err != nil {
-		return utilreconcile.RequeueWith(err)
-	}
-	if res.ShouldStop() {
+	if r.Common.SetFinalizer(&routeMonitor, consts.FinalizerKey) {
+		_, err := r.Common.UpdateReconciledMonitor(&routeMonitor)
+		if err != nil {
+			return utilreconcile.RequeueWith(err)
+		}
```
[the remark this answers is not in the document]
ramonbutter: it's because for most reconcile supplement functions we use a wrapper result as a return value. Meaning the utilreconcile.Result instead of the ctrl.Result. TBH I am also not too big of a fan of it, as I don't understand the gain we got from it. At this point, a translation between the types is happening.
But I believe @georgettica  made some thoughts about it

controllers/clusterurlmonitor/clusterurlmonitor_supplement.go
```diff
@@ -1,368 +1,150 @@
...
-}
-
-func (s *ClusterUrlMonitorSupplement) EnsurePrometheusRuleResourceExists() (utilreconcile.Result, error) {
-	shouldCreate, err, parsedSlo := s.shouldCreatePrometheusRule()
-
-	res, err := s.updateErrorStatus(err)
-	if err != nil || res.RequeueOrStop() {
-		return res, err
+	if s.Common.SetErrorStatus(&clusterUrlMonitor.Status.ErrorStatus, err) {
+		return s.Common.UpdateReconciledMonitorStatus(&clusterUrlMonitor)
```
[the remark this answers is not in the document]
ramonbutter: I changed it to UpdateMonitorResouce

pkg/reconcile/common.go
```diff
@@ -0,0 +1,159 @@
...
+	desiredRef := v1alpha1.NamespacedName{Name: targetNamespace.Name, Namespace: targetNamespace.Namespace}
+	if *reference == (v1alpha1.NamespacedName{}) {
+		*reference = desiredRef
+		return true, nil
+	} else if desiredRef == (v1alpha1.NamespacedName{}) {
+		*reference = desiredRef
+		return true, nil
+	} else if *reference != desiredRef {
+		// TODO Check when this is really required
+		return false, customerrors.InvalidReferenceUpdate
```
ramonbutter, line 86: This case happens when a URLMonitor/RouteMonitor with existing ServiceMinorRef is moved from namespace to namespace. In this case, the reconciler tries to update the existing reference with the ServiceMonitor created in the namespace, but we don't yet support the movement or alterings of already existing sub-resources. 
It's also not really required I would say. In my opinion, we could just create another ServiceMonitor and replace the existing reference. At the moment we are just constantly re-queueing with an error. I am not sure if this is the best way to go.
cc @sulmer-redhat . I think I extracted that part from your code in [here](https://github.com/openshift/route-monitor-operator/blob/e2c7b17d283b95e007f68538bd6b5d7f37e5eec5/controllers/routemonitor/adder/adder.go#L83) Am I wrong about it?

controllers/clusterurlmonitor/clusterurlmonitor.go
```diff
@@ -18,84 +18,112 @@ package clusterurlmonitor
...
 // +kubebuilder:rbac:groups=config.openshift.io,resources=ingresses,verbs=get;list;watch
 // +kubebuilder:rbac:groups=monitoring.coreos.com,resources=prometheusrules,verbs=get;list;watch;create;delete
+// +kubebuilder:rbac:groups=config.openshift.io,resources=clusterversions,verbs=get;list;watch
 
 func (r *ClusterUrlMonitorReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
-	ctx := context.Background()
-	log := r.Log.WithValues("clusterurlmonitor", req.NamespacedName)
+	r.Log.WithValues("clusterurlmonitor", req.NamespacedName)
 
-	clusterUrlMonitor, res, err := r.GetClusterUrlMonitor(req, ctx)
```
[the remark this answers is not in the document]
ramonbutter: good point, that needs definitively to be addressed in the future. For this refactorization, I left the context as part of the struct for now as we are not actually using it. But true best practice recommends having it as the first argument of each function.

pkg/reconcile/common.go
```diff
@@ -0,0 +1,160 @@
...
+	if *reference == (v1alpha1.NamespacedName{}) ||
+		desiredRef == (v1alpha1.NamespacedName{}) {
+		*reference = desiredRef
+		return true, nil
+	} else if *reference != desiredRef {
+		// TODO Check when this is really required
+		return false, customerrors.InvalidReferenceUpdate
+	} else {
+		return false, nil
+	}
```
NautiluX, line 88:
```suggestion
	if *reference == (v1alpha1.NamespacedName{}) ||
		desiredRef == (v1alpha1.NamespacedName{}) {
		*reference = desiredRef
		return true, nil
	}
	if *reference != desiredRef {
		// TODO Check when this is really required
		return false, customerrors.InvalidReferenceUpdate
	}
	return false, nil
```

pkg/servicemonitor/servicemonitor.go
```diff
@@ -0,0 +1,136 @@
...
+		// No similar ServiceMonitor exists
+		if !k8serrors.IsNotFound(err) {
+			return err
+		}
+		return u.Client.Create(u.Ctx, &template)
+	} else if !u.Comparer.DeepEqual(deployedServiceMonitor.Spec, template.Spec) {
+		// Update existing ServiceMonitor for the case that the template changed
+		deployedServiceMonitor.Spec = template.Spec
+		return u.Client.Update(u.Ctx, deployedServiceMonitor)
+	}
```
NautiluX, line 51:
```suggestion
	}
	if !u.Comparer.DeepEqual(deployedServiceMonitor.Spec, template.Spec) {
		// Update existing ServiceMonitor for the case that the template changed
		deployedServiceMonitor.Spec = template.Spec
		return u.Client.Update(u.Ctx, deployedServiceMonitor)
	}
```

pkg/alert/prometheusrule.go
```diff
@@ -0,0 +1,194 @@
...
+		// No similar Prometheus Rule exists
+		if !k8serrors.IsNotFound(err) {
+			return err
+		}
+		return u.Client.Create(u.Ctx, &template)
+	} else if !u.Comparer.DeepEqual(template.Spec, deployedPrometheusRule.Spec) {
+		// Update existing PrometheuesRule for the case that the template changed
+		deployedPrometheusRule.Spec = template.Spec
+		return u.Client.Update(u.Ctx, deployedPrometheusRule)
+	}
```
NautiluX, line 52:
```suggestion
	}
	if !u.Comparer.DeepEqual(template.Spec, deployedPrometheusRule.Spec) {
		// Update existing PrometheuesRule for the case that the template changed
		deployedPrometheusRule.Spec = template.Spec
		return u.Client.Update(u.Ctx, deployedPrometheusRule)
	}
```

NautiluX (commented):
those else ifs should not be necessary, not sure if they are really added here or just moved around, though.
Removing them will at least make @rafael-azevedo happy ;)

PR #119, opened 2021-05-24
feat(labels): add namespace

georgettica:


pkg/util/templates/templates.go
```diff
@@ -100,7 +99,7 @@ func sufficientProbes(windowSize, label string) string {
 }
 
 //render creates a monitoring rule for the defined multiwindow multi-burn rate alert
-func (r *multiWindowMultiBurnAlertRule) render(url, percent, label, alertName string) monitoringv1.Rule {
+func (r *multiWindowMultiBurnAlertRule) render(url, percent, label string, namespacedName types.NamespacedName, alertNameSuffix string) monitoringv1.Rule {
```
NautiluX, line 102: these are a looot of parameters, you think we could move some of them to the rule type?

NautiluX:
/hold
let's fix the build in a separate PR (see #122 )

NautiluX:
we should not rename the URL label in this PR. Let's think about and do it separately, as we need to take into account the impact this will have during the update of the operator on clusters.

pkg/util/templates/templates.go
NautiluX: Glad you mention it! Addressed in the new commit.

georgettica:
sure, I would be glad to fix this issue in a different PR :)

NautiluX (commented):
/lgtm

PR #120, opened 2021-05-27
georgettica/bump slowly to latest operatorsdk [OSD-6289]

georgettica:
currently final verison is 1.13.0


georgettica:
waiting on https://github.com/openshift/release/pull/18918
as it will make the prow things pass

georgettica:
/retest
To see if the new CI works

georgettica:
Seems there are build issue..

Will fix and unhold again

georgettica:
🤞

georgettica:
`/retest`
seems I cannot /draft it while tests are running
I guess
/hold
is in place

georgettica:
I am working on a local branch to get this to work

So far it's going well and soon I will submit a PR for it

2uasimojo:
Note that you *should* be able to `container-make vendor` from any platform and get the right result.

georgettica:
I can test that, and this should work

georgettica:
I want to add some more logging to the hack/teat-integration in order to make  it more readable

georgettica:
partial PR
https://github.com/openshift/route-monitor-operator/pull/137

georgettica:
Seems the code complains on the namespace not existing..
Ideas?

georgettica:
new operator-sdk was introduced, and migration doc shows nothing to be done for our operator

georgettica:
The current issue is the controller-gen is not updating on the ci server

2uasimojo:
> The current issue is the controller-gen is not updating on the ci server

Should it be? According to the [controller-gen target](https://github.com/openshift/route-monitor-operator/pull/120/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R106), it'll only install the binary if there isn't already one in $PATH. But there is, in the [backing image](https://github.com/openshift/boilerplate/blob/328a6a5330ca758439829122dbdfafca9dcc79cf/config/build_image-v2.0.0.sh#L45) CI uses.

If you want this to pass as is, you need to generate your CRDs using that same version (which you could do by running `make generate` from within `quay.io/app-sre/boilerplate:image-v2.1.0`); or change the logic in the `controller-gen` target to override the install if it finds an existing binary at the wrong version.

georgettica:
If by any chance @2uasimojo do you know how to do 
```
If
Else if
Else
```
In makefile that would be great ❤️

2uasimojo:
https://www.gnu.org/software/make/manual/html_node/Conditional-Syntax.html

georgettica:
@2uasimojo seems this was the issue, a space :( https://stackoverflow.com/questions/34761685/extraneous-text-after-else-directive

georgettica:
Fixed that issue by saving the binary locally and added an issue to have a docker image option to it 
Ready for review

controllers/clusterurlmonitor/clusterurlmonitor.go
```diff
@@ -67,16 +63,26 @@ func NewReconciler(mgr manager.Manager, blackboxExporterImage, blackboxExporterN
 	}
 }
 
+const (
+	FinalizerKey string = "clusterurlmonitor.routemonitoroperator.monitoring.openshift.io/finalizer"
+	// PrevFinalizerKey is here until migration to new key is done
+	PrevFinalizerKey string = "clusterurlmonitor.monitoring.openshift.io/clusterurlmonitorcontroller"
```
NautiluX, line 69: Why do we need a different key?

controllers/clusterurlmonitor/clusterurlmonitor_supplement.go
```diff
@@ -110,6 +110,8 @@ func (s *ClusterUrlMonitorReconciler) EnsureMonitorAndDependenciesAbsent(cluster
 	}
 
 	if s.Common.DeleteFinalizer(&clusterUrlMonitor, FinalizerKey) {
+		// ignore the output as we want to remove the PrevFinalizerKey anyways
+		s.Common.DeleteFinalizer(&clusterUrlMonitor, PrevFinalizerKey)
```
NautiluX, line 114: What if we fail to delete the prevfinalizer?

controllers/clusterurlmonitor/clusterurlmonitor_supplement.go
```diff
@@ -118,6 +120,8 @@ func (s *ClusterUrlMonitorReconciler) EnsureMonitorAndDependenciesAbsent(cluster
 
 func (s *ClusterUrlMonitorReconciler) EnsureFinalizerSet(clusterUrlMonitor v1alpha1.ClusterUrlMonitor) (utilreconcile.Result, error) {
 	if s.Common.SetFinalizer(&clusterUrlMonitor, FinalizerKey) {
+		// ignore the output as we want to remove the PrevFinalizerKey anyways
+		s.Common.DeleteFinalizer(&clusterUrlMonitor, PrevFinalizerKey)
```
NautiluX, line 124: What if we fail to delete the prevfinalizer?

controllers/suite_test.go
```diff
@@ -49,33 +49,32 @@ package controllers
...
+//	Expect(k8sClient).NotTo(BeNil())
 //
-//	close(done)
 //}, 60)
 //
 //var _ = AfterSuite(func() {
 //	By("tearing down the test environment")
 //	err := testEnv.Stop()
-//	Expect(err).ToNot(HaveOccurred())
+//	Expect(err).NotTo(HaveOccurred())
```
NautiluX, line 79: Can we remove this code if we don't need it?

main.go
```diff
@@ -58,10 +59,20 @@ func init() {
...
+	flag.BoolVar(&enableLeaderElection, "leader-elect", false,
 		"Enable leader election for controller manager. "+
 			"Enabling this will ensure there is only one active controller manager.")
+	flag.StringVar(&managerConfigPath,
+		"config",
+		"",
+		"The controller will load its initial configuration from this file. "+
+			"Omit this flag to use the default configuration values. "+
+			"Command-line flags override configuration from this file.",
+	)
```
NautiluX, line 75: Is this related to the version upgrade? If not can we separate it into a different PR?

main.go
```diff
@@ -58,10 +59,20 @@ func init() {
 func main() {
 	var metricsAddr string
 	var enableLeaderElection bool
-	flag.StringVar(&metricsAddr, "metrics-addr", ":8080", "The address the metric endpoint binds to.")
-	flag.BoolVar(&enableLeaderElection, "enable-leader-election", false,
+	var probeAddr string
+	var managerConfigPath string
+	flag.StringVar(&metricsAddr, "metrics-bind-address", ":8080", "The address the metric endpoint binds to.")
+	flag.StringVar(&probeAddr, "health-probe-bind-address", ":8081", "The address the probe endpoint binds to.")
```
NautiluX, line 65: Is this related to the version upgrade? If not can we separate it into a different PR?

pkg/util/finalizer/patch_finalizer.go
```diff
@@ -0,0 +1,18 @@
+package finalizer
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func Patch(object metav1.Object, prevFinalizer, finalizer string) {
```
NautiluX, line 7: Is this function used anywhere?

NautiluX (changes requested):
It looks like this PR includes a lot of changes, not all of them seem to be related to the version upgrade. Please at least rebase before merge to structure changes so we don't merge this looong list of commits. It looks like they rather track your work than explain what the changes do.
/approve cancel

controllers/clusterurlmonitor/clusterurlmonitor.go
georgettica: This is a part of the migration, one of the commits talks about it

controllers/suite_test.go
georgettica: It's from operator-sdk, which means it's gonna come back of we want it?

main.go
georgettica: it is related, we are using the config with `config/manager/controller_manager_config.yaml` and to load it we need this new var

main.go
georgettica: this was the reason we did the version upgrade, to have the liveness and readiness

pkg/util/finalizer/patch_finalizer.go
georgettica: nope, it was supposed to and got dismissed (sad because I have tests for it too)

controllers/clusterurlmonitor/clusterurlmonitor_supplement.go
georgettica: the `DeleteFinalizer` is a local command only, so it just removes from an array if the item exists or keeps it empty

also the `s.Common.UpdateMonitorResource` returns the error if something goes sideways

controllers/clusterurlmonitor/clusterurlmonitor_supplement.go
georgettica: same

NautiluX (commented):
/lgtm
/hold
unhold to merge when ready

NautiluX (approved):

georgettica:
@NautiluX so it's ready to unhold?

PR #121, opened 2021-06-01
Jira OSD-6297: Revert alertName parameter in templates.go

Tessg22:
Remove _alertName_ parameter from monitoring rule expression which was causing clusterurlmonitor alert to not fire.


NautiluX (commented):
/lgtm

PR #124, opened 2021-06-03
Add Multiwindow Multi-burn alerts explanation

sulmer-redhat:


georgettica:
seems to explain how we differ with the traditional multiwindowmultiburn

looks fairly straightforward

PR #127, opened 2021-06-08
Updated label

rafael-azevedo:
This PR updates the label RouteMonitorUrl to probeurl. This is done because we want the labels to be in line with current standards per this OSD ticket.

https://issues.redhat.com/browse/OSD-7436

### DIFF
--- a/controllers/routemonitor/adder/adder.go
+++ b/controllers/routemonitor/adder/adder.go
@@ -1,6 +1,8 @@
 package adder
 
 import (
+	"reflect"
+
 	"github.com/go-logr/logr"
 
 	"context"
@@ -51,24 +53,27 @@ func (r *RouteMonitorAdder) EnsureServiceMonitorResourceExists(ctx context.Conte
 
 	namespacedName := types.NamespacedName{Name: routeMonitor.Name, Namespace: routeMonitor.Namespace}
 	resource := &monitoringv1.ServiceMonitor{}
-	populationFunc := func() monitoringv1.ServiceMonitor {
-		return templates.TemplateForServiceMonitorResource(routeMonitor.Status.RouteURL, r.BlackBoxExporterNamespace, namespacedName)
-	}
+	template := templates.TemplateForServiceMonitorResource(routeMonitor.Status.RouteURL, r.BlackBoxExporterNamespace, namespacedName)
 
 	// Does the resource already exist?
-	if err := r.Get(ctx, namespacedName, resource); err != nil {
-		// If this is an unknown error
-		if !k8serrors.IsNotFound(err) {
-			// return unexpectedly
-			return utilreconcile.RequeueReconcileWith(err)
+	err := r.Get(ctx, namespacedName, resource)
+	switch {
+	case err == nil:
+		if !reflect.DeepEqual(template.Spec, resource.Spec) {
+			//Update service monitor if the specs are differentlmao
+			resource.Spec = template.Spec
+			err := r.Update(ctx, resource)
+			if err != nil {
+				utilreconcile.RequeueReconcileWith(err)
+			}
 		}
-		// populate the resource with the template
-		resource := populationFunc()
-		// and create it
-		err = r.Create(ctx, &resource)
+	case k8serrors.IsNotFound(err):
+		err = r.Create(ctx, &template)
 		if err != nil {
 			return utilreconcile.RequeueReconcileWith(err)
 		}
+	case err != nil:
+		return utilreconcile.RequeueReconcileWith(err)
 	}
 
 	//Update status with serviceMonitorRef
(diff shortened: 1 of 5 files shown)


rafael-azevedo:
I checked app-interface and none of the promethusrules have the label RouteMonitorUrl specified . Is there another repo where we could be specifying this label that would trigger an alert ?

RiRa12621:
Shouldn't this be probe_url as per prom standards?

rafael-azevedo:
> Shouldn't this be probe_url as per prom standards?

Ya i've seen short terms not be '_' but I will change it to _ I think in this case it does look better

ramonbutter:
We have one problem that solution as it is as far as I believe:
We compare PrometheusRule changes with deepEqual. So in case, we change the SLOtarget in the RouteMonitor the alert rules are updated. We will because of that also update the PrometheusRules in case we rename the URLlabel in the template, which is good and what we want in the end.
However, we don't have this feature yet for the ServiceMonitor, meaning we just look at whether a ServiceMonitor with that particular name already exists in the target Namespace and don't check for changes in the template. Because many existing clusters in prod have already a ServiceMonitor deployed, they will become outdated. The label in the PrometheusRules would not match with the Labels in the ServiceMonitor anymore, and no alerts would be triggered.
I had the same issue with adding the ClusterID to the ServiceMonitor, which also needs to be updated in existing deployments. I would therefor maybe recommend to wait with merging this PR until I merged my stuff.
I really hope to finish the PR until tomorrow evening.

NautiluX:
> I checked app-interface and none of the promethusrules have the label RouteMonitorUrl specified . Is there another repo where we could be specifying this label that would trigger an alert ?

The Prometheus rule is generated [here](https://github.com/openshift/route-monitor-operator/blob/7e175770d2015cdf6a200accefd923bbcb921a08/pkg/util/templates/templates.go#L137), the ServiceMonitor [here](https://github.com/openshift/route-monitor-operator/blob/7e175770d2015cdf6a200accefd923bbcb921a08/pkg/util/templates/templates.go#L23). The PrometheusRule filters on the label [here](https://github.com/openshift/route-monitor-operator/blob/7e175770d2015cdf6a200accefd923bbcb921a08/pkg/util/templates/templates.go#L104).

The PrometheusRule will be updated when deployed with your PR, as RMO checks if it needs to be updated [here](https://github.com/openshift/route-monitor-operator/blob/7e175770d2015cdf6a200accefd923bbcb921a08/controllers/routemonitor/routemonitor_supplement.go#L110) but we don't have such a check for the ServiceMonitor (neither [for the RouteMonitor](https://github.com/openshift/route-monitor-operator/blob/a32c9b3a0c0e39798deee7195fd72de8c55f7956/controllers/routemonitor/adder/adder.go#L61) nor the [ClusterUrlMonitor](https://github.com/openshift/route-monitor-operator/blob/f797d9bedd94cc8db3293054adc5a242655778b2/controllers/clusterurlmonitor/clusterurlmonitor_supplement.go#L82)), so the labels in the PrometheusRule won't match the labels in the ServiceMonitor, and hence the metrics, anymore after deploying to existing clusters.

rafael-azevedo:
Fixed bug in the comparison on routermonitor_supplement and added comparison of the service monitor

controllers/routemonitor/adder/adder.go
```diff
@@ -51,23 +53,28 @@ func (r *RouteMonitorAdder) EnsureServiceMonitorResourceExists(ctx context.Conte
...
+			}
 		}
-		// populate the resource with the template
-		resource := populationFunc()
-		// and create it
-		err = r.Create(ctx, &resource)
+
+		// return unexpectedly
+		return utilreconcile.RequeueReconcileWith(err)
+	}
```
NautiluX, line 70: this code changes the logic, I don't think that's on purpose.

georgettica:
Enjoy the
/Lgtm
But I will
/Hold
Just in case someone else wants to look and approve

controllers/routemonitor/adder/adder_test.go
```diff
@@ -88,9 +90,9 @@ var _ = Describe("Adder", func() {
 			Return(update.ErrorResponse).
 			Times(update.CalledTimes)
 
-		mockClient.EXPECT().Get(gomock.Any(), gomock.Any(), gomock.Any()).
-			Return(get.ErrorResponse).
-			Times(get.CalledTimes)
+		// mockClient.EXPECT().Get(gomock.Any(), gomock.Any(), gomock.Any()).
+		// 	Return(get.ErrorResponse).
+		// 	Times(get.CalledTimes)
```
NautiluX, line 95: If we don't need this, can we remove it?

.tours/rmo-intro.tour
```diff
@@ -0,0 +1,80 @@
+{
```
NautiluX, line 1: where is this file coming from? I think this should already be on master, probably need to rebase?

NautiluX (commented):
/lgtm

PR #128, opened 2021-06-10
change base domain from configv1/ingress to configv1/dns

sulmer-redhat:


NautiluX (changes requested):
I'm pretty sure you need to adjust the integration tests as well

NautiluX (approved):
/lgtm

PR #132, opened 2021-06-16
fix API clusterurlmonitor sample

sulmer-redhat:


georgettica:
why not, it's based on the conversation and it works better then +1

/lgtm

PR #133, opened 2021-06-23
OSD-7377: removed client/fake usage in tests and added some more tests

ramonbutter:


controllers/routemonitor/routemonitor_test.go
```diff
@@ -389,26 +386,15 @@ var _ = Describe("Routemonitor", func() {
 	//--------------------------------------------------------------------------------------
 	Describe("GetRoute", func() {
 
-		f := fuzz.New()
```
ramonbutter, line 392: fuzzy testing is here not required IMO as we don't have any functions that actually process the name and namespace strings. So there is no need to try crazy string values. 
We just forward the strings to the Kubernetes client in order to fetch the CR. Any error caused by this should be handled via the client and is simply returned as the client error (same as NotFoundError). However, in our unit tests, we are anyways completely mocking this functinality so we can drop this.

georgettica: @RiRa12621 sad to see it go but what is said is true

maybe we can card an option to make fuzzy testing somewhere else?

georgettica (commented):
seems good
/lgtm

controllers/routemonitor/routemonitor_test.go
ramonbutter: I think its still used for other tests. I was just looking into this very particular case

NautiluX:
OSD-7377

PR #137, opened 2021-07-04
feat: make test-integration more robust

georgettica:
- make flags be multiple choise
- when build fails, show it's log
- TODO: when a deploy fails, show it's log

### DIFF
--- a/hack/test-integration.sh
+++ b/hack/test-integration.sh
@@ -7,14 +7,65 @@ export NAMESPACE=${NAMESPACE:-openshift-route-monitor-operator}
 export IMAGE_NAME=route-monitor-operator
 export KUBECTL=oc
 
-SKIP=${1:-""}
+
+function parseArgs {
+  SKIP_BUILD=${SKIP_BUILD:-}
+  SKIP_CLEANUP=${SKIP_CLEANUP:-}
+  SKIP_TEST=${SKIP_TEST:-}
+
+  PARSED_ARGUMENTS=$(getopt -o 'n:' --long 'namespace:,kubeconfig:,skip-test,skip-build,skip-cleanup' -- "$@")
+  eval set -- "$PARSED_ARGUMENTS"
+  while :
+  do
+    case "$1" in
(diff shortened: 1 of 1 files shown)


georgettica:
Please look at the teat-integration logs and decide it it looks good

You can also look at the script

hack/test-integration.sh
```diff
@@ -44,6 +97,13 @@ function waitForDeployment {
     return 1
   fi
 }
+function verifyPodIsSuccessfull {
+	podName=$(oc -n "$NAMESPACE" get po -lapp="route-monitor-operator" -ojsonpath='{.items[0].metadata.name}')
+	echo "status of the pod"
+	oc -n "$NAMESPACE" get po "$podName" -ojsonpath='{.status}'
+	echo "pod logs"
+	oc -n "$NAMESPACE" logs "$podName"
```
ramonbutter, line 105: We could maybe pipe it to in the end.
| egrep "^|error|warning"
Not important at all, but why not

hack/test-integration.sh
```diff
@@ -65,18 +125,23 @@ function cleanup {
...
 deployOperator
 
 waitForDeployment
+verifyPodIsSuccessfull
 
-if [[ "$SKIP" != "--skip-test" ]]; then
+if [[ -z $SKIP_TEST ]]; then
   runTests
+  verifyPodIsSuccessfull
 fi
```
ramonbutter, line 147: nit, I would maybe just use one verifyPodIsSuccessfull
in the end, that gets always called. Otherwise, we have duplicated log lines.
The naming is maybe also a bit confusing, since it is not verifying anything but printing the pod status

ramonbutter (commented):
lgtm

hack/test-integration.sh
georgettica: not gonna fix as there might be more interesting info in the logs

ramonbutter: yes this egrep statement doesn't filter but only highlights the error and warning statements.
But doesn't matter. I am fine with how it is

georgettica:
/hold
we are forcing CI to rerun a failing test

PR #139, opened 2021-07-20
Upgrade BP and CRD versions

sam-nguyen7:
Upgrade boilerplate version to latest
Upgrade crd to v1


2uasimojo:
The bp update and the CRD transition are pretty well unrelated for rmo, but given the existing [overrides](https://github.com/openshift/route-monitor-operator/blob/2f78e2d9ad7814a42c2da4b97df0a7ad6df9f5f9/Makefile#L11), it shouldn't matter.

The v1 CRDs install cleanly, and `oc explain` works, so

/lgtm

PR #141, opened 2021-08-06
OSD-7984: use ubi-minimal as base image

ramonbutter:
There is not more to do, it just works: 
Checked it here:
https://quay.io/repository/rbutter/route-monitor-operator/manifest/sha256:f880d65a7a07c7d89bc50898ec46930acd4f5d18a9c878c490e2c34deed807b5?tab=vulnerabilities


georgettica:
is there a size increase?

ramonbutter:
yep unpacked its 145 MB vs 44 MB atm. In quay packed it's 54MB vs 17MB. But I guess its still no super big issue

georgettica:
/lgtm
if the test-integration passes this is good enogh for me, can be pushed to proc too

PR #147, opened 2021-10-14
switch image to quay.io/prometheus/blackbox-exporter

maorfr:
quay.io/app-sre/prom-blackbox-exporter is going away. use the official image instead (same image).


georgettica:
/hold
so you won't need to wait for my stuff.. currently RMO merged should wait until I fix the `./hack/app_sre_build_deploy.sh` (currently 3 PR's in the works)

georgettica:
/unhold
tested my old change and it seems to pass, the only ones that are not updated have syncsets disabled

PR #152, opened 2021-12-08
feat(blackbox): add deployment self-heal

georgettica:
this will allow to update the deployment if needed


pkg/blackboxexporter/blackboxexporter.go
```diff
@@ -80,11 +89,17 @@ func (b *BlackBoxExporter) EnsureBlackBoxExporterDeploymentExists() error {
...
-		}
+		return b.Client.Create(b.Ctx, &resource)
+	}
+
+	template := populationFunc()
+	if !b.Comparer.DeepEqual(resource.Spec, template.Spec) {
+		// template differs from resource, update to reconcile
+		b.Log.Info("the spec of the resource", "resource.Spec", resource.Spec)
+		resource.Spec = template.Spec
+		return b.Client.Update(b.Ctx, &resource)
```
georgettica, line 100: this fails because the resource has changed

georgettica:
checks failed as were interrupted

pkg/blackboxexporter/blackboxexporter.go
georgettica: also, it will always be different after the first creation as there are default controllers that update deployments

georgettica:
@rafael-azevedo I wrote the code and Don remember the state, but there is a good base for the issue..
I hope this can get pulled in so I can make it work

georgettica:
as it's super stale and not happening. will close :(

PR #155, opened 2022-01-13
Add operator icon

NautiluX:


georgettica:
Have you tested this locally?
If so than
/lgtm

But wating for your response so
/hold

georgettica:
/retest
Succeed locally

PR #157, opened 2022-01-26
chore(github-action): fix local build

georgettica:
the local build would use `podman` and that would fail on an issue I didn't want to tackle

moving to `docker` fixed it


georgettica:
this solves the local  `github action` failure

PR #161, opened 2022-02-15
feat(routemonitor): force https if exists

georgettica:
this change forces routemonitors to use https if they can

Additional things
---
- add gitignore to config.bak
- fix static analysis stuff (whitespace in comments / comparison to empty object)
---

the actual change is in https://github.com/openshift/route-monitor-operator/pull/161/files#diff-fd106a311406a33d76b3a5ab0c6cd44d936a8835991a06f26efd655ebde2b63aR197-R200


georgettica:
this failed locally as the servicemonitors didn't create.

I know that the blackbox-exporter can recieve this string and works well

georgettica:
related to OSD-9959

ramonbutter:
and why was the service monitor not created for you locally? 
But as the integration tests are working fine, and I don't see any mistake in the logic.
Would have a lgtm

georgettica:
It was because the pulled url and the compared url were different, so it didn't go to the next step

controllers/suite_test.go
```diff
@@ -16,7 +16,7 @@ limitations under the License.
 
 package controllers
 
-//import (
```
shibumi, line 19: @georgettica wonder if it makes sense to remove this file completely if we do not use it. What do you think? Can we remove this savely? In the worst case, we still have it in our git history.

georgettica: yeah, this can happen no problem.
my idea was to make `operator-sdk` be able to regenerate and update it's generated data.

but it's a dream I am not stiving on making, so NP

NautiluX (commented):
/lgtm

georgettica:
❤️

PR #165, opened 2022-05-02
[OSD-10503] build(Dockerfile): update base image to ubi-micro

Tessg22:
This change is to make sure there are no critical or high vulnerabilities in base image


georgettica:
Also @Tessg22 this is not really usable as this will not allow running a shell, which is required for tekton to run this container

georgettica:
oh, it's RMO.
go ahead

PR #166, opened 2022-05-03
OSD-10503: build(Dockerfile): update ubi-micro image tag

Tessg22:
Update ubi-micro image tag from latest to versioned.

### DIFF
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -21,7 +21,7 @@ RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 GO111MODULE=on go build -mod vendor -o
 
 # Use distroless as minimal base image to package the manager binary
 # Refer to https://github.com/GoogleContainerTools/distroless for more details
-FROM registry.access.redhat.com/ubi8/ubi-micro:latest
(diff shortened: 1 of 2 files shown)


georgettica:
/lgtm
/approve
another step in the right direction

Nikokolas3270:
Are we sure we want we want to get rid of `latest`?
As said here, using `latest` is widely accepted & used:
https://coreos.slack.com/archives/CCX9DB894/p1651506047185529?thread_ts=1651504202.651549&cid=CCX9DB894
As discussed yesterday, fixing a version for a dependency also has it issues if we forget to bump that version from time to time.
I am not against using a fixed version, but then, for the sake of consistency, it would be good to encourage the practice elsewhere as well.

georgettica:
I guess using latest is some sort of automation, but we should have a process of rolling the image across the fleet / checking the digests of the images on the fleet to make sure we are actually on top of this problem

shibumi:
I agree with @georgettica  here.. having `latest` is nice for the sake of automation, but we shouldn't sacrifice deterministic behavior for it.

PR #170, opened 2022-05-23
fix(pipeline.dockerfile): revert base image change

Tessg22:
Due to lack of microdnf package in ubi-micro image, build pipeline is currently failing. Hence this PR.
Will follow up by checking if there can be workaround for required packages such as make, which and gcc. Then we could use ubi-micro image.


georgettica:
`latest`?

georgettica:
as it's a revert, less going to be stringent.

next PR should check on this
/approve
/lgtm

georgettica:
Oof, fix and ping me again

PR #171, opened 2022-05-26
Enable FIPS build

wanghaoran1988:


2uasimojo:
Coming here from https://github.com/openshift/boilerplate/pull/228, which smelled like kind of a hacky solution to the test-integration failure.

Turns out that `oc start-build` will by default exclude `.git` directories when it copies files from the `--from-dir` or `--from-repo`; but you can override that behavior by specifying an explicit `--exclude`. Specifically, to exclude nothing (thus including `.git`) you can say `--exclude=""`. I reproduced the failure locally, then added that and `configure-fips.sh` succeeded.

```diff
diff --git a/hack/test-integration.sh b/hack/test-integration.sh
index 661252e..22f29cc 100755
--- a/hack/test-integration.sh
+++ b/hack/test-integration.sh
@@ -55,7 +55,9 @@ function buildImage {
   echo -e "\n\nSTARTING BUILD\n\n"
   oc adm new-project "$NAMESPACE" || true
   oc new-build --binary --strategy=docker --name "$IMAGE_NAME" -n "$NAMESPACE" || true
-  oc start-build -n "$NAMESPACE" "$IMAGE_NAME" --from-dir . -F
+  # .git dirs are excluded by default. Use --exclude="" to include them,
+  # as the build relies on discovering git-isms from the repo.
+  oc start-build -n "$NAMESPACE" "$IMAGE_NAME" --from-dir . --exclude="" -F
  
   oc set image-lookup -n "$NAMESPACE" "$IMAGE_NAME"
 }
```

PR #179, opened 2022-09-22
[OSD-11801]update dependencies and boilerplate to match with other operators

bmeng:
Update to golang 1.17 and kube related dependencies to 0.24.3
Make some Make file change to compatible with boilerplate

There are 3 commits in the PR, the remove vendor one can be ignored as it is big and just removing the vendor dir from the repo


mrbarge:
Attempting to do a `make build-push` to generate the registry catalog image as well as the operator image failed:

```
Traceback (most recent call last):
  File "./boilerplate/openshift/golang-osd-operator/csv-generate/common-generate-operator-bundle.py", line 106, in <module>
    with open('config/templates/csv-template.yaml', 'r') as stream:
FileNotFoundError: [Errno 2] No such file or directory: 'config/templates/csv-template.yaml'
```

Potentially missing this file?

bmeng:
The `config/templates/csv-template.yaml` was not there originally, how would it build image with app-sre pipeline? @NautiluX do you have any idea?

mrbarge:
> The `config/templates/csv-template.yaml` was not there originally, how would it build image with app-sre pipeline? @NautiluX do you have any idea?

As noted elsewhere [0] currently the CI pipeline performs a `make build-push` and is suffering from the same issue. Previously, it had been building it via a `hack/app_sre_create_image_catalog.sh` script.

As part of bringing this operator back into a properly-boilerplated model, a goal should be to migrate it to boilerplate's native approach to building the catalog image (or spike the exploration of that)

[0] https://coreos.slack.com/archives/CFJD1NZFT/p1664275784837149

mrbarge:
I have given this a pretty thorough test (the write-up is in OSD-11801) and as a result I think this is good to go.
Removing the hold.
/unhold

mrbarge:
@NautiluX thanks for that update, that seems to be testing well.

Except one catch. boilerplate doesn't like the presence of `openshift-route-monitor-operator.Namespace.yaml` and treats that as an unsupported resource. List of supported resources is defined here [0]

Options are:
- We exclude/remove the Namespace file from the manifest generation
- We look at adding `Namespace` to boilerplate's supported resources [0] (*caveat: see below).

For what it's worth, I tried the latter option and regenerated the catalog with the namespace included. OLM did not like it and refused to install the operator. So I am not sure if option 2 is really a viable one.

[0] https://github.com/mrbarge/boilerplate/blob/master/boilerplate/openshift/golang-osd-operator/csv-generate/common-generate-operator-bundle.py#L24

mrbarge:
I have tested/verified that the latest commit:
- builds a catalog via boilerplate successfully, based on manifests built by `make manifests`
- deploys/updates on a cluster via OLM successfully

PR #181, opened 2022-10-11
update image tags to patch CVE's

tonytheleg:
For [OSD-13308](https://issues.redhat.com/browse/OSD-13308) and [OSD-13325](https://issues.redhat.com/browse/OSD-13325)
- updates boilerplate to `a6c27570590858412f03ce4b84e7d6bb1ecfdc7a` to ensure the ubi-micro version used in the registry image is updated to patch CVE's required for FedRAMP initial audit
- updates the ubi-micro version used for the operator to the same value version as version `8.6-484` still has the CVE present that we are working to patch

I've tested both the registry image and operator image in a FedRAMP cluster: registry image properly deploys the operator and there are no errors in either the registry or operator pod logs. I was also able to build the images using `make build-push` which is what the AppSRE pipeline will do too. Thanks!


NautiluX (commented):
/lgtm

2uasimojo (commented):
Drive by lgtm

PR #182, opened 2022-10-24
Supporting of rhobs.monitoring resources

anispate:
[OSD-13437](https://issues.redhat.com/browse/OSD-13437)

As an OSD Operator for Hypershift, I want my operator to generate servicemonitors.rhobs.monitoring when deployed to Hypershift. Therefore, When an operator is deployed to hypershift, a ServiceMonitor from rhobs.monitoring is generated as opposed to a monitoring.coreos.com ServiceMonitor.

Removed GetServiceMonitor function as it was unused.


controllers/interfaces.go
```diff
@@ -59,6 +59,12 @@ type ServiceMonitorHandler interface {
 
 	// DeleteServiceMonitorDeployment deletes a ServiceMonitor refrenced by a namespaced name
 	DeleteServiceMonitorDeployment(serviceMonitorRef v1alpha1.NamespacedName) error
+
```
mjlshen, line 62: TODO: Add descriptive comments

anispate: Added Descriptions

fahlmant:
@anispate can you please add a description to the PR?

pkg/servicemonitor/servicemonitor.go
```diff
@@ -53,6 +54,26 @@ func (u *ServiceMonitor) UpdateServiceMonitorDeployment(template monitoringv1.Se
 	return nil
 }
 
+// Creates or Updates Service Monitor Deployment according to the template if enable of the hypershift
+func (u *ServiceMonitor) Hypershift_UpdateServiceMonitorDeployment(template rhobsv1.ServiceMonitor) error {
```
NautiluX, line 58: I don't like that we need to duplicate the whole function just to use a different API group. Is there a smarter way to achieve this?

NautiluX, line 58:
```suggestion
func (u *ServiceMonitor) HypershiftUpdateServiceMonitorDeployment(template rhobsv1.ServiceMonitor) error {
```

NautiluX: Is it really depending on hypershift? Or is it more for users who use OO and RHOBS CRDs instead of CMO?

pkg/servicemonitor/servicemonitor.go
```diff
@@ -73,13 +94,41 @@ func (u *ServiceMonitor) DeleteServiceMonitorDeployment(serviceMonitorRef v1alph
 	return u.Client.Delete(u.Ctx, resource)
 }
 
+// Deletes the ServiceMonitor Deployment for HyperShift
+func (u *ServiceMonitor) HyperShift_DeleteServiceMonitorDeployment(serviceMonitorRef v1alpha1.NamespacedName) error {
```
NautiluX, line 98:
```suggestion
func (u *ServiceMonitor) HyperShiftDeleteServiceMonitorDeployment(serviceMonitorRef v1alpha1.NamespacedName) error {
```

NautiluX (changes requested):
This PR adds a lot of code duplication. Is there probably a different way to achieve this if the only difference is the API group? Also, I'm wondering if "enabling hypershift" is the right terminology, or if we make it depending on the resource types? Like... "enabling rhobs resource definitions"?

pkg/servicemonitor/servicemonitor.go
mjlshen: Agreed, but any ideas? I'm not aware of a better way to do this at least, especially if these two API groups start diverging in the future.

rafael-azevedo (commented):
Another General note Function names should be camel Case and should not include '_' or '-'

pkg/servicemonitor/servicemonitor.go
rafael-azevedo: @anispate Do you think there is any better way here instead of a whole function to duplicate the api group ? Should we consider maybe outputting a structure that can hold several API groups as supported and have one function that knows how to create this ?

mjlshen: API groups are types though, what we would like is:

"If HyperShift, then call `func UpdateServiceMonitorDeployment(template *rhobsv1)`, else `func UpdateServiceMonitorDeployment(template *monitoringv1)`, but I don't think this is possible.

controllers/routemonitor/routemonitor.go
```diff
@@ -60,6 +60,7 @@ func NewReconciler(mgr manager.Manager, blackboxExporterImage, blackboxExporterN
 		ServiceMonitor:   servicemonitor.NewServiceMonitor(ctx, client),
 		Prom:             alert.NewPrometheusRule(ctx, client),
 		Common:           reconcileCommon.NewMonitorResourceCommon(ctx, client),
+		Hypershift:       enablehypershift,
```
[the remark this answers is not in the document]
rafael-azevedo: We should change the spelling here to the recommended camel case from Manuel

controllers/routemonitor/routemonitor_supplement.go
```diff
@@ -64,10 +64,18 @@ func (r *RouteMonitorReconciler) EnsureServiceMonitorExists(routeMonitor v1alpha
 
 	// update ServiceMonitor if requiredctrl
 	namespacedName := types.NamespacedName{Name: routeMonitor.Name, Namespace: routeMonitor.Namespace}
-	serviceMonitorTemplate := servicemonitor.TemplateForServiceMonitorResource(routeMonitor.Status.RouteURL, r.BlackBoxExporter.GetBlackBoxExporterNamespace(), namespacedName, r.Common.GetClusterID())
-	err := r.ServiceMonitor.UpdateServiceMonitorDeployment(serviceMonitorTemplate)
-	if err != nil {
-		return utilreconcile.RequeueReconcileWith(err)
+	if r.Hypershift == true {
+		serviceMonitorTemplate := servicemonitor.HyperShift_TemplateForServiceMonitorResource(routeMonitor.Status.RouteURL, r.BlackBoxExporter.GetBlackBoxExporterNamespace(), namespacedName, r.Common.GetClusterID())
+		err := r.ServiceMonitor.Hypershift_UpdateServiceMonitorDeployment(serviceMonitorTemplate)
```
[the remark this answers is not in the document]
rafael-azevedo: @anispate I this has not been addressed yet. I see we changed that variable name, but I think we should wrap this logic here in one function that knows how to handle both hypershift and non hypershift I'd be happy to have a call to discuss this if necessary

pkg/servicemonitor/servicemonitor.go
```diff
@@ -33,6 +34,22 @@ const (
...
 )
 
+func (u *ServiceMonitor) TemplateAndUpdateServiceMonitorDeployment(routeURL, blackBoxExporterNamespace string, namespacedName types.NamespacedName, clusterID string) error {
+	params := map[string][]string{
+		// Currently we only support `http_2xx` as module
+		"module": {"http_2xx"},
+		"target": {routeURL},
+	}
+
+	if u.Hypershift == true {
```
rafael-azevedo, line 44: You can remove the else here

```suggestion

	if u.Hypershift == true {
		s := u.HyperShiftTemplateForServiceMonitorResource(routeURL, blackBoxExporterNamespace, params, namespacedName, clusterID)
		return u.HypershiftUpdateServiceMonitorDeployment(s)
	} 
	s := u.TemplateForServiceMonitorResource(routeURL, blackBoxExporterNamespace, params, namespacedName, clusterID)
	return u.UpdateServiceMonitorDeployment(s)
	
}
```

pkg/servicemonitor/servicemonitor.go
```diff
@@ -53,52 +70,67 @@ func (u *ServiceMonitor) UpdateServiceMonitorDeployment(template monitoringv1.Se
...
+			}
+			// Resource doesn't exist, nothing to do
+			return nil
+		}
 
-	// Currently we only support `http_2xx` as module
-	// Still make it a variable so we can easily add functionality later
-	modules := []string{"http_2xx"}
+		return u.Client.Delete(u.Ctx, resource)
+	} else {
```
rafael-azevedo, line 114: This else can be removed liked the above example since there is a return contained in the if statement above

pkg/servicemonitor/servicemonitor.go
rafael-azevedo: Also we should only have one updateservicemoinitor deployment function. Similar to how you made DeleteServiceMonitorDeployment work for both Hypershift and non hypershift

mjlshen: It's not possible to have one `TemplateServiceMonitor` nor `UpdateServiceMonitor` function because it interacts with a rhobsv1/monitoringv1 resource directly via its input/outputs.

`DeleteServiceMonitor` did not, which made it possible. Chaining `TemplateServiceMonitor` and `UpdateServiceMonitor` together enables this one function.

mjlshen: i.e. `TemplateServiceMonitor` has non-relevant inputs, but outputs a ServiceMonitor
`UpdateServiceMonitor` takes a ServiceMonitor as an input and has non-relevant outputs
-->
`TemplateAndUpdateServiceMonitor` has non-relevant inputs/outputs, so it can hold logic inside it to pick the right servicemonitor

pkg/servicemonitor/servicemonitor.go
```diff
@@ -33,6 +34,22 @@ const (
...
+	params := map[string][]string{
+		// Currently we only support `http_2xx` as module
+		"module": {"http_2xx"},
+		"target": {routeURL},
+	}
+
+	if u.Hypershift == true {
+		s := u.HyperShiftTemplateForServiceMonitorResource(routeURL, blackBoxExporterNamespace, params, namespacedName, clusterID)
+		return u.HypershiftUpdateServiceMonitorDeployment(s)
+	} else {
```
rafael-azevedo, line 47: else not necessary here

PR #183, opened 2022-10-31
OSD-12884 get the cluster domain from the infrastructure APIServerURL

yithian:
this will give a correct url for hypershift clusters as well as classic-style rosa/osd clusters

### DIFF
--- a/hack/test-integration.sh
+++ b/hack/test-integration.sh
@@ -61,6 +61,7 @@ function buildImage {
 }
 
 function verifyForBuildSuccess {
+  sleep 300
   local latestJobName phase
   latestJobName="$IMAGE_NAME"-$( oc -n "$NAMESPACE" get buildconfig "$IMAGE_NAME" -ojsonpath='{.status.lastVersion}') 
   phase=$(oc -n "$NAMESPACE" get build "$latestJobName" -ojsonpath='{.status.phase}')
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -42,6 +42,14 @@ rules:
   - list
   - watch
 - apiGroups:
+  - config.openshift.io
+  resources:
+  - infrastructures
+  verbs:
+  - get
+  - list
+  - watch
+- apiGroups:
   - monitoring.coreos.com
   resources:
   - prometheusrules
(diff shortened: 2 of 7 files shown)


yithian:
/hold

this has yet to be tested on a hypershift cluster since i haven't been able to create one yet

boranx (commented):
lgtm

one note: do we need to re-generate the crds? I think they already exist on https://github.com/openshift/route-monitor-operator/blob/master/deploy/clusterurlmonitors.monitoring.openshift.io.CustomResourceDefinition.yaml

yithian:
I rebased off of master and ~cherry-picked only the actual code and RBAC changes into a new branch and force-pushed it here. Hopefully that'll unjam whatever CRD stuff is going on.

yithian:
@boranx I _finally_ got the tests to all pass. Please re-review at your convenience!

mjlshen (approved):
/lgtm

hack/test-integration.sh
```diff
@@ -61,6 +61,7 @@ function buildImage {
 }
 
 function verifyForBuildSuccess {
+  sleep 300
```
boranx, line 64: was this put on purpose? why do we need this?

boranx (commented):
/lgtm

I wanted to highlight that [OSD-12884](https://issues.redhat.com/browse/OSD-12884) was created to enable rmo to be deployed onto hosted cluster. If we need rmo to be deployed on managementcluster, there might be other tweaks needed too (like route CRs need to be queriable from mgmt cluster for each hosted cluster)
If we move with hosted cluster, the deployment of it can be done using acm policies, similar to https://github.com/openshift/managed-cluster-config/pull/1447

hack/test-integration.sh
yithian: That was needed because otherwise, the integration test would expect builds to complete instantly and would fail with `build was not completed fully, the state was Running but expected to be 'Complete'`. I added this sleep statement to allow the builds time to complete

yithian:
> /lgtm
> 
> I wanted to highlight that [OSD-12884](https://issues.redhat.com/browse/OSD-12884) was created to enable rmo to be deployed onto hosted cluster. If we need rmo to be deployed on managementcluster, there might be other tweaks needed too (like route CRs need to be queriable from mgmt cluster for each hosted cluster) If we move with hosted cluster, the deployment of it can be done using acm policies, similar to [openshift/managed-cluster-config#1447](https://github.com/openshift/managed-cluster-config/pull/1447)

My understanding is that RMO will need to be deployed onto both hosted clusters (to monitor the console) and management clusters (to monitor the hosted clusters' API)

I've been informed that deployment of operators to hosted clusters will need to be done with packaging-operator, so that's what I'll be working on next

boranx:
ack, overall looks good to me, feel free to merge & test on staging

yithian:
@NautiluX @Tessg22 @ninabauer @rafael-azevedo can I get an /approve on this?

PR #188, opened 2023-01-12
Revert olm update

sam-nguyen7:
management clusters are classic osd clusters and therefore already have RMO deployed. PR created to revert previous olm template update.


rafael-azevedo (approved):
/lgtm

PR #200, opened 2023-03-17
log into quay before pushing the package image

yithian:
this prevents the package image push from failing due to lack of credentials


mjlshen (commented):
/lgtm

PR #203, opened 2023-04-06
OSD-15737 pass hypershift flag to servicemonitor

yithian:
\<servicemonitor\>.Hypershift is what determines whether a servicemonitor.monitoring.rhobs resource is created instead of a servicemonitor.monitoring.coreos.com resource

### DIFF
--- a/pkg/servicemonitor/servicemonitor.go
+++ b/pkg/servicemonitor/servicemonitor.go
@@ -23,9 +23,10 @@ type ServiceMonitor struct {
 
 func NewServiceMonitor(ctx context.Context, c client.Client, enablehypershift bool) *ServiceMonitor {
 	return &ServiceMonitor{
-		Client:   c,
-		Ctx:      ctx,
-		Comparer: &util.ResourceComparer{},
+		Client:     c,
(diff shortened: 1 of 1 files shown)


pkg/servicemonitor/servicemonitor.go
```diff
@@ -23,9 +23,10 @@ type ServiceMonitor struct {
 
 func NewServiceMonitor(ctx context.Context, c client.Client, enablehypershift bool) *ServiceMonitor {
 	return &ServiceMonitor{
-		Client:   c,
-		Ctx:      ctx,
-		Comparer: &util.ResourceComparer{},
+		Client:     c,
+		Ctx:        ctx,
+		Comparer:   &util.ResourceComparer{},
+		Hypershift: enablehypershift,
```
sam-nguyen7, line 29: How is the `enableHypershift` determined? This should only be true if RMO is reconciling clusterurlmonitors deployed on HCP namespaces right?

sam-nguyen7:
Implemented in https://github.com/openshift/route-monitor-operator/pull/204

PR #204, opened 2023-04-10
[WIP] OSD-15834 - Create rhobs servicemonitor for HCP clusterurlmonitors

tnierman:
https://issues.redhat.com/browse/OSD-15834

---

TODO


controllers/routemonitor/routemonitor_supplement.go
```diff
@@ -63,7 +63,8 @@ func (r *RouteMonitorReconciler) EnsureServiceMonitorExists(routeMonitor v1alpha
 
 	// update ServiceMonitor if requiredctrl
 	namespacedName := types.NamespacedName{Name: routeMonitor.Name, Namespace: routeMonitor.Namespace}
-	if err := r.ServiceMonitor.TemplateAndUpdateServiceMonitorDeployment(routeMonitor.Status.RouteURL, r.BlackBoxExporter.GetBlackBoxExporterNamespace(), namespacedName, r.Common.GetClusterID()); err != nil {
+	isHCP := false
+	if err := r.ServiceMonitor.TemplateAndUpdateServiceMonitorDeployment(routeMonitor.Status.RouteURL, r.BlackBoxExporter.GetBlackBoxExporterNamespace(), namespacedName, r.Common.GetClusterID(), isHCP); err != nil {
```
sam-nguyen7, line 67: `r.Common.GetClusterID()` returns mgmt clusterID. Can we pass the cluster of HCP in here if isHCP := true

pkg/reconcile/common.go
```diff
@@ -140,16 +141,37 @@ func (u *MonitorResourceCommon) UpdateMonitorResourceStatus(cr client.Object) (r
...
-		}
-		u.ClusterID = string(version.Spec.ClusterID)
+// GetOSDClusterID returns the ID for the cluster based on its ClusterVersion
+func (u *MonitorResourceCommon) GetOSDClusterID() (string, error) {
+	var version configv1.ClusterVersion
+	err := u.Client.Get(u.Ctx, client.ObjectKey{Name: "version"}, &version)
+	if err != nil {
+		return "", err
 	}
-	return u.ClusterID
```
tnierman, line 152: Caching the cluster ID probably made sense when RMO only acted on a single cluster, but now that it's being adapted for HyperShift and potentially acting on different clusters with each reconcile, we should avoid caching metadata between executions.

config/rbac/role.yaml
```diff
@@ -128,6 +128,16 @@ rules:
   - get
   - patch
   - update
+- apiGroups:
```
sam-nguyen7, line 131: This can be removed as the changes are contained in PR #205 so we can look into merging that in first

deploy/route-monitor-operator-manager-role.ClusterRole.yaml
```diff
@@ -130,6 +130,16 @@ rules:
       - get
       - patch
       - update
+  - apiGroups:
```
sam-nguyen7, line 133: Can remove change in favor of #205

pkg/reconcile/common.go
```diff
@@ -140,16 +141,37 @@ func (u *MonitorResourceCommon) UpdateMonitorResourceStatus(cr client.Object) (r
...
+	hcp, err := u.GetHCP(ns)
+	if err != nil {
+		return "", err
+	}
+	return hcp.Spec.ClusterID, nil
+}
+
+// GetHCP returns the HostedControlPlane object in the namespace provided. If more than one HCP object exists in the same namespace, an error is returned
+func (u *MonitorResourceCommon) GetHCP(ns string) (hypershiftv1beta1.HostedControlPlane, error) {
+	// Retrieve the HostedControlPlane in order to lookup the associated hostedCluster object
```
sam-nguyen7, line 165: Can implement backoff retries here in the future as HCP may not exist for some time

tnierman: To be clear, are you suggesting this be added now, or just making a note to add this in the future?

config/rbac/role.yaml
tnierman: The code is autogenerated, so lets hang onto it for now. 

If #205 gets merged first, a simple rebase should clear up this diff. If this gets merged first, the logic here won't represent a "partial" solution and instead contain everything needed to get clusterurlmonitors working for hypershift :slightly_smiling_face:

deploy/route-monitor-operator-manager-role.ClusterRole.yaml
tnierman: See [above](https://github.com/openshift/route-monitor-operator/pull/204#discussion_r1164738886)

pkg/reconcile/common.go
sam-nguyen7: Just making note to add this in the next iteration

sam-nguyen7 (commented):
/lgtm
/approve

PR #213, opened 2023-07-14
Update clusterurlmonitor to use the private NLB domain for private HCPs

bdmiller3:
Currently clusterurlmonitor attempts to use a public endpoint for private clusters, causing both inaccurate and unnecessary alerts.

Closes https://issues.redhat.com/browse/OSD-17069

### DIFF
--- a/controllers/clusterurlmonitor/clusterurlmonitor_supplement.go
+++ b/controllers/clusterurlmonitor/clusterurlmonitor_supplement.go
@@ -12,6 +12,7 @@ import (
 	"github.com/openshift/route-monitor-operator/pkg/alert"
 	blackboxexporterconsts "github.com/openshift/route-monitor-operator/pkg/consts/blackboxexporter"
 	utilreconcile "github.com/openshift/route-monitor-operator/pkg/util/reconcile"
+	corev1 "k8s.io/api/core/v1"
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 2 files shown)


controllers/clusterurlmonitor/clusterurlmonitor_supplement.go
```diff
@@ -214,13 +215,27 @@ func (s *ClusterUrlMonitorReconciler) getHypershiftClusterDomain(monitor v1alpha
...
+
+	// In the case of private clusters, the domain will be the ingress of the private router
+	if endpointAccess == hypershiftv1beta1.Private {
+		service := corev1.Service{}
+		err := s.Client.Get(s.Ctx, types.NamespacedName{Name: "private-router", Namespace: monitor.Namespace}, &service)
+		if err != nil {
+			return "", fmt.Errorf("could not retrieve private router in namespace '%s'; Reason: %s", monitor.Namespace, err.Error())
+		}
+
+		ingress := service.Status.LoadBalancer.Ingress[0].Hostname
```
mjlshen, line 234: During the initial creation of the service, this status field may not exist yet as it takes some time for a Load Balancer to be created in AWS, so we should add a couple `nil` guards around this and return an error if the service is not fully populated yet.

It should be safe to assume that `service.Status` exists, but anything after may not when this code runs (eventually we can assume that it will be populated).

mjlshen (approved):
/lgtm

PR #216, opened 2023-07-18
OSD-15543 - Update the package template to deploy ACM policies to 'openshift-operator-lifecycle-manager'

tnierman:
Updates the hypershift package template to deploy ACM policies to the `openshift-operator-lifecycle-manager` namespace, as the current `openshift-acm-policies` namespace no longer exists


mjlshen (approved):
/lgtm

PR #217, opened 2023-07-18
Revert #216 and deploy ACM policies to service-clusters instead of mgmt clusters

tnierman:
Reverts the change to the selectorsyncset deployment namespace made by #216 and selects `service-cluster` `clusterdeployments` instead


mjlshen (approved):
/lgtm

PR #219, opened 2023-07-19
Update all HCP clusterurlmonitors to check the internal LB

bdmiller3:
Previously this only applied to private clusters' service monitors.


mjlshen (approved):
/lgtm

PR #222, opened 2023-07-21
Reduce abstraction in the main reconcile loop

mjlshen:
This commit's goal is to make the main reconcile loop look like:

```go
func (r *ClusterUrlMonitorReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
	r.log = ctrllog.FromContext(ctx).WithName("controller").WithName("ClusterUrlMonitor")

	clusterUrlMonitor := new(v1alpha1.ClusterUrlMonitor)
	if err := r.Get(ctx, req.NamespacedName, clusterUrlMonitor); err != nil {
		return ctrl.Result{}, client.IgnoreNotFound(err)
	}

	if clusterUrlMonitor.ObjectMeta.DeletionTimestamp.IsZero() {
		// It's not being deleted, so if it does not have our finalizer, then lets add the finalizer and update the object.
	} else {
		// It's not deleting, do stuff
	}
	
	return ctrl.Result{}, nil
}
```

By removing calls in the main reconcile loop to the internal packages `utilreconcile` and `finalizer`. Future PRs will remove those two internal packages completely. It also remove contexts that were stored in structs.

[OSD-17568](https://issues.redhat.com//browse/OSD-17568)


mjlshen:
/hold
For reviews

PR #207, opened 2023-04-24
Add .spec.skipPrometheusRule to the RouteMonitor CRD in order to... 

mjlshen:
ensure no PrometheusRule exists when configured.

[OSD-16127](https://issues.redhat.com//browse/OSD-16127)

This PR does this by adding a new field which will cause the controller to delete existing PrometheusRules that are mentioned in `.status.PrometheusRuleRef`, otherwise if `.status.PrometheusRuleRef` is empty, just skip the creation.

PR #223, opened 2023-07-21
Add .spec.skipPrometheusRule to the ClusterURLMonitor CRD in order to…

mjlshen:
… ensure no PrometheusRule exists when configured

Very similar to #207 where this was done to the RouteMonitor CRD

[OSD-16950](https://issues.redhat.com//browse/OSD-16950)

### DIFF
--- a/controllers/clusterurlmonitor/clusterurlmonitor_supplement.go
+++ b/controllers/clusterurlmonitor/clusterurlmonitor_supplement.go
@@ -25,6 +25,20 @@ const (
 
 // Takes care that right PrometheusRules for the defined ClusterURLMonitor are in place
 func (s *ClusterUrlMonitorReconciler) EnsurePrometheusRuleExists(clusterUrlMonitor v1alpha1.ClusterUrlMonitor) (utilreconcile.Result, error) {
+	// If .spec.skipPrometheusRule is true, ensure that the PrometheusRule does NOT exist
+	if clusterUrlMonitor.Spec.SkipPrometheusRule {
+		// Cleanup any existing PrometheusRules and update the status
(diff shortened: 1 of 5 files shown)


controllers/clusterurlmonitor/clusterurlmonitor_supplement.go
```diff
@@ -25,6 +25,20 @@ const (
 
 // Takes care that right PrometheusRules for the defined ClusterURLMonitor are in place
 func (s *ClusterUrlMonitorReconciler) EnsurePrometheusRuleExists(clusterUrlMonitor v1alpha1.ClusterUrlMonitor) (utilreconcile.Result, error) {
+	// If .spec.skipPrometheusRule is true, ensure that the PrometheusRule does NOT exist
+	if clusterUrlMonitor.Spec.SkipPrometheusRule {
```
yithian, line 29: Will `SkipPrometheusRule` have a default value? Or is it expected to be set on all clusterUrlMonitors?

If there's not a default value, it might be good to confirm that there's a `SkipPrometheusRule` present in the Spec to avoid a null pointer exception

mjlshen: It is! It's defaulted as false https://github.com/openshift/route-monitor-operator/pull/223/files#diff-1f765eef49a09245f809b9f25819cb33df4c1be9421a3f735cdb96d8104efe80R41-R42

yithian: Sorry, I overlooked the kubebuilder stuff and only looked at the json definition!

/lgtm

PR #225, opened 2023-07-24
Refactor - Reducing Complexity in RMO

mjlshen:
This PR is a combination of commits to remove unnecessary interfaces, mocks, and ginkgo tests to replace them with vanilla Go table-driven tests and utilizing upstream Kubernetes client mocks. It will admittedly be difficult to review as a whole and may make more sense for this PR to be split up into many PRs (let me know), I would recommend reviewing the changes a commit-at-a-time.

Furthermore, it removes contexts from being stored in structs and moves them to become the first parameter in functions when they are needed as is general practice https://go.dev/blog/context

[OSD-17568](https://issues.redhat.com//browse/OSD-17568)


pkg/alert/prometheusrule_test.go
```diff
@@ -1,207 +1,119 @@
...
 
-		deepEqual = ResourceComparerMockHelper{}
-		get = testhelper.MockHelper{}
-		create = testhelper.MockHelper{}
-		update = testhelper.MockHelper{}
-		delete = testhelper.MockHelper{}
-
-		prometheusRuleRef = v1alpha1.NamespacedName{}
-		prometheusRule = monitoringv1.PrometheusRule{}
+	var namespacedName = types.NamespacedName{
```
NautiluX, line 25: +1 to simplify unit tests. However, the rest of the code base uses BDD tests with Gomega assertions. Why do you think this specific file would benefit from table tests using `assert.` assertions? Should we use Ginkgo table tests instead?

pkg/blackboxexporter/blackboxexporter.go
```diff
@@ -1,121 +1,105 @@
...
 	}
 	for i := range clusterUrlMonitors.Items {
 		objectsDependingOnExporter = append(objectsDependingOnExporter, &clusterUrlMonitors.Items[i])
 	}
-	b.Log.V(4).Info("Number of objects depending on BlackBoxExporter:", "amountOfObjects", len(objectsDependingOnExporter))
 
-	if len(objectsDependingOnExporter) == 1 && finalizer.WasDeleteRequested(objectsDependingOnExporter[0]) {
-		b.Log.V(3).Info("Deleting BlackBoxResources: decided to clean BlackBoxExporter resources")
-		return blackboxexporter.DeleteBlackBoxExporter, nil
+	// If there's only one clusterurlmonitor/routemonitor left and it's deleting, delete the blackboxexporter
```
NautiluX, line 57: Is there a way to express this statement in code?

pkg/servicemonitor/servicemonitor_test.go
```diff
@@ -1,207 +1,193 @@
...
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
 )
 
-type ResourceComparerMockHelper struct {
-	CalledTimes int
-	ReturnValue bool
+func TestTemplateAndUpdateServiceMonitorDeployment(t *testing.T) {
```
NautiluX, line 18: +1 to simplify unit tests. However, the rest of the code base uses BDD tests with Gomega assertions. Why do you think this specific file would benefit from table tests using `assert.` assertions? Should we use Ginkgo table tests instead?

NautiluX (commented):
/lgtm

PR #241, opened 2023-08-22
add hosted cluster monitors to the RMO package

yithian:
this will deploy the api and console monitors to hosted clusters as part of the package (instead of relying on an ACM policy in managed-cluster-config to deploy them)


mjlshen (approved):
/lgtm

[OSD-17687](https://issues.redhat.com//browse/OSD-17687)

PR #254, opened 2023-12-01
[OSD-19634] Schedule BlackBoxExporter pods on control-plane if cluster version is >= 4.14 and it's a privatelink

zmird-r:
Ticket: [OSD-19634](https://issues.redhat.com/browse/OSD-19634)


mjlshen (approved):
/lgtm

PR #268, opened 2024-02-29
Adds '--enable-hypershift' flag to the hypershift PKO package

tnierman:
https://issues.redhat.com/browse/OSD-21000

---

Adds an additional flag `--enable-hypershift=true` to the `package` Make target, which is only applied to HyperShift management clusters. This will enable the newly introduced `hostedcontrolplane` controller on management clusters while leaving deployments on classic OSD/ROSA unchanged.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -180,7 +180,7 @@ endif
 
 .PHONY: package
 package: kubectl-package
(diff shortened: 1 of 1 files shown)


NautiluX (commented):
/lgtm

PR #272, opened 2024-03-08
Utilize HCP cluster ID when running on HyperShift

tnierman:
https://issues.redhat.com/browse/OSD-21000

Based on the branch for #271

---

Updates the routemonitor controller's ID retrieval logic when running on HyperShift, so that the `_id` label on it's `probe_success` metric refers to the hosted cluster and not the management cluster. 

As a result, the alerts generated from the controller will more clearly indicate which cluster requires attention.

### DIFF
--- a/controllers/routemonitor/routemonitor_supplement.go
+++ b/controllers/routemonitor/routemonitor_supplement.go
@@ -1,6 +1,7 @@
 package routemonitor
 
 import (
+	"context"
 	"errors"
 	"fmt"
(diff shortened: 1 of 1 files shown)


tnierman:
Here's a better diff of these changes, without the reverts from #271: https://github.com/tnierman/route-monitor-operator/compare/revert-270-osd-21000...tnierman:route-monitor-operator:fix_hcp_cluster_id

PR #274, opened 2024-03-14
OSD-21605 - Only deploy PKO package to IBM infrastructure

tnierman:
https://issues.redhat.com/browse/OSD-21605

---

Ensures that the PKO package is only deployed to IBM infrastructure. This way, IBM HCP clusters' kube-apiservers will still be monitored from the data-plane, while all other clusters will be monitored from the management plane


controllers/hostedcontrolplane/hostedcontrolplane.go
```diff
@@ -225,7 +225,7 @@ func (r *HostedControlPlaneReconciler) buildInternalMonitoringRouteMonitor(route
 			Route: v1alpha1.RouteMonitorRouteSpec{
 				Name:      route.Name,
 				Namespace: route.Namespace,
-				Port:      6443,
+				Port:      int64(hostedcontrolplane.Status.ControlPlaneEndpoint.Port),
```
mjlshen, line 228: Is this is problematic if the status field isn't populated yet (since status fields aren't guaranteed to be populated)?

mjlshen (approved):
/lgtm

typeid (commented):
@tnierman / @mjlshen I understand this is the straightforward fix to the issue we're currently facing. However, it has the downside of requiring us to keep fleet wide monitoring for data plane metrics (ClusterMetricsMissing). 

We can merge this to fix staging, but we should look for a better solution for prod IMO. Could we maybe get the port from the kube-apiserver service in the HCP namespace? Alternatively, extract the version from the hostedcluster object and use 443 or 6443 based on that?

iamkirkbater:
/lgtm

@typeid this monitoring solution is just a stop gap until we have better monitoring at the edge,which won't be dependent on RMO.

PR #277, opened 2024-03-19
Discover API server port from service

NautiluX:


controllers/hostedcontrolplane/hostedcontrolplane.go
```diff
@@ -150,8 +151,19 @@ func (r *HostedControlPlaneReconciler) deployInternalMonitoringObjects(ctx conte
 		}
 	}
 
+	// Quick fix to discover the API server port from the service resource
+	apiServerService := v1.Service{}
+	err = r.Client.Get(ctx, types.NamespacedName{Name: "kube-apiserver", Namespace: hostedcontrolplane.Namespace}, &apiServerService)
+	if err != nil {
+		return fmt.Errorf("couldn't query API server service resource: %v", err)
```
typeid, line 158:
```suggestion
		return fmt.Errorf("couldn't query API server service resource: %w", err)
```

typeid (commented):
Makes sense to me, small nit with the %wrap vs %value for consistency

PR #279, opened 2024-03-26
Add unit tests to hcp controller

tnierman:
https://issues.redhat.com/browse/OSD-21346

---

Adds initial suite of unit tests to the new hostedcontrolplane controller, as well as a few fixes to the controller itself which were surfaced when running the `lint` target.


mjlshen (approved):
/lgtm

PR #282, opened 2024-04-16
OSD-22380 - Ensure HCPs are not reconciled when not Ready

tnierman:
https://issues.redhat.com/browse/OSD-22380

---

Fixes the HCP readiness check to ensure HostedControlPlane objects that are not in the Ready state are not reconciled.

### DIFF
--- a/controllers/hostedcontrolplane/hostedcontrolplane.go
+++ b/controllers/hostedcontrolplane/hostedcontrolplane.go
@@ -86,7 +86,7 @@ func (r *HostedControlPlaneReconciler) Reconcile(ctx context.Context, req ctrl.R
 	if err != nil {
 		if kerr.IsNotFound(err) {
 			log.Info("HostedControlPlane not found, assumed deleted")
-			return utilreconcile.RequeueWith(nil)
+			return utilreconcile.Stop()
 		}
 		log.Error(err, "unable to fetch HostedControlPlane")
 		return utilreconcile.RequeueWith(err)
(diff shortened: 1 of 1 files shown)


controllers/hostedcontrolplane/hostedcontrolplane.go
```diff
@@ -116,7 +119,7 @@ func (r *HostedControlPlaneReconciler) Reconcile(ctx context.Context, req ctrl.R
 	// Check if the HostedControlPlane is ready
 	if !hostedcontrolplane.Status.Ready {
 		log.Info("skipped deploying monitoring objects: HostedControlPlane not ready")
-		utilreconcile.RequeueWith(nil)
+		return utilreconcile.RequeueWith(nil)
```
mjlshen, line 122: Passing `nil` here is misleading - https://github.com/openshift/route-monitor-operator/blob/master/pkg/util/reconcile/reconcile_wrapper.go#L25-L31 it doesn't actually trigger a reconcile. I'm not sure about this whole controller, but specifically here I think we want to pass something like this instead
```suggestion
		return utilreconcile.RequeueWith(errors.New("skipped deploying monitoring objects: HostedControlPlane not ready"))
```

mjlshen (approved):
/lgtm

mjlshen (commented):
/label tide/merge-method-squash

mjlshen (approved):
/lgtm

PR #289, opened 2024-05-03
[OSD-20249] Create and delete dynatrace http monitors

hbhushan3:
Hostedcontrolplane controller will create and delete HTTP monitors in response to the creation or deletion of public hostedcluster resources so that we can measure the availability of the HCP's kube-apiserver.

Ticket - https://issues.redhat.com/browse/OSD-20249


controllers/hostedcontrolplane/hostedcontrolplane.go
```diff
@@ -329,3 +381,288 @@ func (r *HostedControlPlaneReconciler) SetupWithManager(mgr ctrl.Manager) error
...
+		if loc["name"] == locationName && loc["type"] == "PUBLIC" && loc["cloudPlatform"] == "AMAZON_EC2" {
+			return loc["entityId"].(string), nil
+		}
+	}
+
+	return "", fmt.Errorf("location '%s' not found", locationName)
+}
+
+func (APIClient *APIClient) createDynatraceHTTPMonitor(monitorName, apiUrl, clusterId, dynatraceEquivalentClusterRegionId string) (string, error) {
+	monitorConfig := map[string]interface{}{
```
iamkirkbater, line 512: Seeing all of these `map[string]interface{}` definitions is a smell to me.

Would it maybe make more sense to use this as a [template](https://pkg.go.dev/text/template) if we need some of the more advanced functions?  Or could we define this as its own struct if we need to interact with it on a more granular level?

hbhushan3: We are using template and structs now. Thanks!

controllers/hostedcontrolplane/hostedcontrolplane.go
```diff
@@ -323,3 +458,277 @@ func (r *HostedControlPlaneReconciler) SetupWithManager(mgr ctrl.Manager) error
...
+		return nil, err
+	}
+
+	req.Header.Set("Authorization", "Api-Token "+APIClient.apiToken)
+	req.Header.Set("Content-Type", "application/json")
+
+	return APIClient.httpClient.Do(req)
+}
+
+func (r *HostedControlPlaneReconciler) getSecret(ctx context.Context) (string, string, error) {
```
iamkirkbater, line 482: I think this function name is also a bit too generic. Potentially we could rename this to something like `getDynatraceSecrets` or something?

controllers/hostedcontrolplane/hostedcontrolplane.go
```diff
@@ -323,3 +458,277 @@ func (r *HostedControlPlaneReconciler) SetupWithManager(mgr ctrl.Manager) error
...
+	hostedcontrolplane.SetLabels(labels)
+
+	err := r.Client.Update(ctx, hostedcontrolplane)
+	if err != nil {
+		return fmt.Errorf("error updating hostedcontrolplane monitor: %v", err)
+	}
+	return nil
+}
+
+func (r *HostedControlPlaneReconciler) GetAPIServerHostname(hostedcontrolplane *hypershiftv1beta1.HostedControlPlane, log logr.Logger) (string, error) {
```
iamkirkbater, line 532: Does this need to be a function on the reconciler? I don't see `r` used anywhere within here.

Similarly with the logger - do we need this if we're not using it?

controllers/hostedcontrolplane/hostedcontrolplane.go
```diff
@@ -323,3 +458,277 @@ func (r *HostedControlPlaneReconciler) SetupWithManager(mgr ctrl.Manager) error
...
+	}
+	if len(valueBytesDynatraceTenant) == 0 {
+		return "", "", fmt.Errorf("%s is empty", dynatraceTenantKey)
+	}
+	valueDynatraceTenant := string(valueBytesDynatraceTenant)
+
+	return valueDynatraceApiToken, valueDynatraceTenant, nil
+}
+
+func (APIClient *APIClient) getDynatraceHTTPMonitorID(ctx context.Context, log logr.Logger, hostedcontrolplane *hypershiftv1beta1.HostedControlPlane) (string, error) {
```
iamkirkbater, line 511: Is there a reason why `ctx` and `log` are passed in here?

controllers/hostedcontrolplane/hostedcontrolplane.go
```diff
@@ -323,3 +458,277 @@ func (r *HostedControlPlaneReconciler) SetupWithManager(mgr ctrl.Manager) error
...
+func (r *HostedControlPlaneReconciler) GetAPIServerHostname(hostedcontrolplane *hypershiftv1beta1.HostedControlPlane, log logr.Logger) (string, error) {
+	for _, service := range hostedcontrolplane.Spec.Services {
+		if service.Service == "APIServer" {
+			return service.ServicePublishingStrategy.Route.Hostname, nil
+		}
+	}
+	return "", fmt.Errorf("APIServer service not found in the hostedcontrolplane")
+}
+
+func (APIClient *APIClient) getDynatraceEquivalentClusterRegionId(hostedcontrolplane *hypershiftv1beta1.HostedControlPlane) (string, error) {
```
iamkirkbater, line 541: I think it might be better to try to keep `hostedcontrolplane` objects as decoupled from `dynatrace` objects as much as possible. In this function I see that we're only using the `hostedcontrolplane.Spec.Platform.AWS.Region` out of the object. How would you feel about just taking a string as an argument and letting the caller of this function handle figuring out how to get the string it needs.  This will result in less-coupled code but also would make this easier to test.

controllers/hostedcontrolplane/hostedcontrolplane.go
```diff
@@ -323,3 +458,277 @@ func (r *HostedControlPlaneReconciler) SetupWithManager(mgr ctrl.Manager) error
...
+	for _, loc := range locationResponse.Locations {
+		if loc.Name == locationName && loc.Type == "PUBLIC" && loc.CloudPlatform == "AMAZON_EC2" {
+			return loc.EntityID, nil
+		}
+	}
+
+	return "", fmt.Errorf("location '%s' not found", locationName)
+}
+
+func (APIClient *APIClient) createDynatraceHTTPMonitor(monitorName, apiUrl, clusterId, dynatraceEquivalentClusterRegionId string) (string, error) {
```
iamkirkbater, line 605: If we're renaming `APIClient` to be something like `DynatraceAPIClient` then it might make sense to name some of these functions that are on the struct things like `createHTTPMonitor`.

iamkirkbater (changes requested):
Overall I think this looks great at a high level. I didn't dive too deep into any given function, though, and will have to take a second pass at it. 

I have a few small changes around naming that I'd like to see. While I would like to see the refactoring work happen I don't think that any of the refactoring work I'd like to see should block anything from merging.

hbhushan3:
Refactored and updated PR here - https://github.com/openshift/route-monitor-operator/pull/311

PR #291, opened 2024-05-13
OSD-20859: fix: Add support for custom console domains in Blackbox Exporter

Tessg22:
This PR fixes a bug in the Blackbox Exporter that prevented it from handling custom console domains. Tested with dummy URLs in the staging cluster.

Jira: [OSD-20859](https://issues.redhat.com//browse/OSD-20859)


controllers/routemonitor/routemonitor_supplement.go
```diff
@@ -235,7 +236,24 @@ func (r *RouteMonitorReconciler) EnsureRouteURLExists(route routev1.Route, route
...
-		r.Log.V(3).Info("RouteURL mismatch: currentRouteURL and extractedRouteURL are not equal, taking extractedRouteURL as source of truth")
+		client := &http.Client{
+			CheckRedirect: func(req *http.Request, via []*http.Request) error {
+				return http.ErrUseLastResponse
+			},
+		}
+
+		resp, err := client.Head(extractedRouteURL)
+		if err != nil {
+			r.Log.V(3).Info("Failed to make HEAD request to the URL")
```
xiaoyu74, line 247: Looks the requests fails only logs an `info` message, it would be better we can improve the error handling for this HEAD request to be more visible for debugging purposes. WDYT?

controllers/routemonitor/routemonitor_supplement.go
```diff
@@ -235,7 +236,24 @@ func (r *RouteMonitorReconciler) EnsureRouteURLExists(route routev1.Route, route
...
+		}
+
+		if resp != nil && (resp.StatusCode == http.StatusMovedPermanently || resp.StatusCode == http.StatusFound || resp.StatusCode == http.StatusSeeOther || resp.StatusCode == http.StatusTemporaryRedirect || resp.StatusCode == http.StatusPermanentRedirect) {
```
xiaoyu74, line 250: Looks the newly added logic hasn't got unit tests covering, especially the redirection handling part? It would be better we can have test cases accordingly.

image

controllers/routemonitor/routemonitor_supplement.go
```diff
@@ -235,7 +236,24 @@ func (r *RouteMonitorReconciler) EnsureRouteURLExists(route routev1.Route, route
...
+				return http.ErrUseLastResponse
+			},
+		}
+
+		resp, err := client.Head(extractedRouteURL)
+		if err != nil {
+			r.Log.V(3).Error(err, "Failed to make HEAD request to the URL")
+		}
+
+		if resp != nil && (resp.StatusCode == http.StatusMovedPermanently || resp.StatusCode == http.StatusFound || resp.StatusCode == http.StatusSeeOther || resp.StatusCode == http.StatusTemporaryRedirect || resp.StatusCode == http.StatusPermanentRedirect) {
```
bmeng, line 250: Would it be easier we just check the return code matches regex `3xx`?

controllers/routemonitor/routemonitor_supplement.go
```diff
@@ -235,7 +236,24 @@ func (r *RouteMonitorReconciler) EnsureRouteURLExists(route routev1.Route, route
...
+			r.Log.V(3).Info("RouteURL mismatch: redirection status code received, taking defaultRouteURL as source of truth")
+			extractedRouteURL = resp.Header.Get("Location")
+			defer resp.Body.Close()
+		} else {
+			r.Log.V(3).Info("RouteURL mismatch: currentRouteURL and extractedRouteURL are not equal, taking extractedRouteURL as source of truth")
```
bmeng, line 255: In which condition, it will go to this else?

controllers/routemonitor/routemonitor_supplement.go
Tessg22: My understanding is 300 and 304 aren't exact fit for redirection. https://pkg.go.dev/net/http#Client.Get

controllers/routemonitor/routemonitor_supplement.go
Tessg22: If the route.Status.Ingress[0].Host doesn't have any redirects.

Tessg22:
@bmeng amd @xiaoyu74, should we merge this?

mjlshen (changes requested):
There could be a problem in the original bug report as it claims that the blackbox-exporter returns
```
tls: failed to verify certificate: x509: certificate signed by unknown authority" 
```
while `curl -kIL` successfully connects - however `curl -k` ignores SSL validation? Are we sure we have a bug with the blackbox exporters? Commented on the card as well.

hectorakemp (commented):
The fix `lgtm` -  I guess the only reservation would be @mjlshen's one - should we just ignore SSL verification on console routes instead? See https://issues.redhat.com/browse/OSD-20859?focusedId=24833762&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-24833762 for more details.

controllers/routemonitor/routemonitor_supplement.go
```diff
@@ -235,7 +236,24 @@ func (r *RouteMonitorReconciler) EnsureRouteURLExists(route routev1.Route, route
...
+
+		resp, err := client.Head(extractedRouteURL)
+		if err != nil {
+			r.Log.V(3).Error(err, "Failed to make HEAD request to the URL")
+		}
+
+		if resp != nil && (resp.StatusCode == http.StatusMovedPermanently || resp.StatusCode == http.StatusFound || resp.StatusCode == http.StatusSeeOther || resp.StatusCode == http.StatusTemporaryRedirect || resp.StatusCode == http.StatusPermanentRedirect) {
+			r.Log.V(3).Info("RouteURL mismatch: redirection status code received, taking defaultRouteURL as source of truth")
+			extractedRouteURL = resp.Header.Get("Location")
+			defer resp.Body.Close()
```
tnierman, line 253: Should this defer be moved to outside the if/else block, so the response object is always closed? Or do we want to not close if we receive a bad response for some reason?

controllers/routemonitor/routemonitor_supplement.go
```diff
@@ -254,3 +260,24 @@ func (r *RouteMonitorReconciler) getHostedControlPlane(namespace string) (hypers
 	}
 	return hcpList.Items[0], nil
 }
+
+// checkRedirect makes a HEAD request to the provided URL and handles any redirection responses.
+// If a redirection status code (300-399) is received, it updates the routeMonitor object's InsecureSkipTLSVerify field to true
+// and sets the extractedRouteURL to the redirection location.
```
NautiluX, line 266: what does setting the extractedRouteURL mean? As far as I see this function only updates the `InsecureSkipTLSVerify` field on the routeMonitor?

Tessg22: The purpose of checkRedirect() is to also update the extractedRouteURL value. The code has been updated accordingly.

controllers/routemonitor/routemonitor_supplement.go
```diff
@@ -234,11 +234,17 @@ func (r *RouteMonitorReconciler) EnsureRouteURLExists(route routev1.Route, route
...
 
 	if currentRouteURL != "" && extractedRouteURL != currentRouteURL {
-		r.Log.V(3).Info("RouteURL mismatch: currentRouteURL and extractedRouteURL are not equal, taking extractedRouteURL as source of truth")
+		err := r.checkRedirect(extractedRouteURL, routeMonitor)
+		if err != nil {
+			r.Log.V(3).Error(err, "RouteURL mismatch: failed to check redirects")
+			return utilreconcile.RequeueReconcileWith(err)
+		} else {
+			r.Log.V(3).Info("RouteURL mismatch: currentRouteURL and extractedRouteURL are not equal, taking extractedRouteURL as source of truth")
+		}
```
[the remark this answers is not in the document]
Tessg22: This would be incorrect. Since if the redirection status code is received, defaultRouteURL is the source of truth.

controllers/routemonitor/routemonitor_supplement.go
```diff
@@ -234,11 +234,17 @@ func (r *RouteMonitorReconciler) EnsureRouteURLExists(route routev1.Route, route
 	}
 
 	if currentRouteURL != "" && extractedRouteURL != currentRouteURL {
-		r.Log.V(3).Info("RouteURL mismatch: currentRouteURL and extractedRouteURL are not equal, taking extractedRouteURL as source of truth")
+		err := r.checkRedirect(extractedRouteURL, routeMonitor)
+		if err != nil {
+			r.Log.V(3).Error(err, "RouteURL mismatch: failed to check redirects")
+			return utilreconcile.RequeueReconcileWith(err)
+		} else {
+			r.Log.V(3).Info("RouteURL mismatch: currentRouteURL and extractedRouteURL are not equal, taking extractedRouteURL as source of truth")
```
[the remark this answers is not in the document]
Tessg22: I am sorry, I don't understand your comment. There are two states in which either defaultRouteURL or extractedRouteURL are the source of truth.

controllers/routemonitor/routemonitor_supplement.go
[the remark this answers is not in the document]
NautiluX: I think it should be the same condition (`err == nil`) since there is a return in the if branch?

controllers/routemonitor/routemonitor_supplement.go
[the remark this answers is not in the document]
NautiluX: that's because it doesn't make sense the way I wrote it :D

So when `checkRedirect` doesn't return an error, and finds a redirect, it will log
```
RouteURL mismatch: redirection status code received, taking defaultRouteURL as source of truth
```
in line 278. Since it didn't return an error, this else-branch will be evaluated and log
```
RouteURL mismatch: currentRouteURL and extractedRouteURL are not equal, taking extractedRouteURL as source of truth
```
which seems confusing to me?

controllers/routemonitor/routemonitor_supplement.go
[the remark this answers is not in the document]
Tessg22: You are right!

PR #302, opened 2024-06-27
remove achvatal from owners

yithian:
:saluting_face:


mjlshen (commented):
/close

Merged #308 instead

PR #310, opened 2024-07-31
[OSD-23678] generate the console-ErrorBudgetBurn alerts only when the console is default route

bmeng:
Screenshot from 2024-07-30 11-02-40

As the metrics above shown, there is a built-in metric with name `console_url`, which will add a new entry everytime when the custom console route added/updated, and set the value to 1 only for the active console route. It can be used to identify if the console route is customized.

With above, append the following condition to the existing console-ErrorBudgetBurn alert

```
and count(console_url{url="$DEFAULT_CONSOLE_URL"} == 1) > 0
```
So it will alert SRE only when the active console_url is the default console route.

### DIFF
--- a/pkg/alert/prometheusrule.go
+++ b/pkg/alert/prometheusrule.go
@@ -4,16 +4,18 @@ import (
 	"context"
 	"fmt"
 	"strconv"
+	"strings"
 	"time"
 
 	prometheus "github.com/prometheus/common/model"
 
-	"github.com/openshift/route-monitor-operator/api/v1alpha1"
-	util "github.com/openshift/route-monitor-operator/pkg/reconcile"
-	"github.com/openshift/route-monitor-operator/pkg/servicemonitor"
 	monitoringv1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1"
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/openshift/route-monitor-operator/api/v1alpha1"
+	util "github.com/openshift/route-monitor-operator/pkg/reconcile"
+	"github.com/openshift/route-monitor-operator/pkg/servicemonitor"
 
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/intstr"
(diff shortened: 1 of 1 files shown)


pkg/alert/prometheusrule.go
```diff
@@ -117,6 +119,12 @@ func (r *multiWindowMultiBurnAlertRule) render(url string, percent string, names
 		" and " +
 		sufficientProbes(r.longWindow, labelSelector)
 
+	if namespacedName.Name == "console" {
+		defaultConsoleURL := strings.TrimSuffix(url, "/health")
```
xiaoyu74, line 123: How do we normally determine the console URL is default or custom? Just checking if it's in the `console` namespace?

xiaoyu74:
With the current updated logic to only check the default active console URL, it looks that we are completely ignoring situations where requests fail to reach the custom console URL (e.g redirecting to a custom console URL and got certificate issues). Is that expected?

pkg/alert/prometheusrule.go
bmeng: the `url` here comes from the `routeMonitor.Status.RouteURL` which was set to the default and I did not see it  would change

bmeng:
> With the current updated logic to only check the default active console URL, it looks that we are completely ignoring situations where requests fail to reach the custom console URL (e.g redirecting to a custom console URL and got certificate issues). Is that expected?

yes, you can check the card for more details, but the goal is to monitor the console when it is not customized only

xiaoyu74:
> > With the current updated logic to only check the default active console URL, it looks that we are completely ignoring situations where requests fail to reach the custom console URL (e.g redirecting to a custom console URL and got certificate issues). Is that expected?
> 
> yes, you can check the card for more details, but the goal is to monitor the console when it is not customized only

oh, I was thinking if we should have customer getting notifications or alerts when using custom console URL and got issues. Just checked the card, looks you've already discussed and that makes sense to me.
```
With that said, if the customer modified the console, that means the SLO does not make sense to us anymore, as it is not fully controlled by us.
If we feel the existing ingress operator or console operator cannot do monitoring them properly, we can enhance that part with OCP team.
```

pkg/alert/prometheusrule.go
xiaoyu74: Sorry, I am still not quite clear how's the `url` comes from the `routeMonitor.Status.RouteURL` from the function calls. Do you mean it's derived from this struct?
```
// RouteMonitorStatus defines the observed state of RouteMonitor
type RouteMonitorStatus struct {
	// RouteURL is the url extracted from the Route resource
	RouteURL          string         `json:"routeURL,omitempty"`
	ServiceMonitorRef NamespacedName `json:"serviceMonitorRef,omitempty"`
	PrometheusRuleRef NamespacedName `json:"prometheusRuleRef,omitempty"`
	ErrorStatus       string         `json:"errorStatus,omitempty"`
}
```
But how's `RouteURL` above is determiend as the default URL? Is that explicitly defined something like `RouteURL: "https://console-openshift-console.apps.example.com/health`?

bmeng: The route fetched by the other function `GetRoute()` in routemonitor controller by reading the the route with name `console` in openshift-console ns, and append the suffix ("/health" here)

mjlshen (approved):
Not ideal, this is kind of an unexpected change from the "api" of route-monitor-operator, but it seems to be the simplest way to move forward.

/lgtm

PR #313, opened 2024-08-21
Support Private HCP API Monitoring

hbhushan3:
Adds support for Creation and deletion of HTTP monitors in response to the creation or deletion of private hostedcluster resources so that we can measure the availability of the HCP's kube-apiserver.

Ticket - https://issues.redhat.com/browse/OSD-20250


controllers/hostedcontrolplane/hostedcontrolplane.go
```diff
@@ -431,6 +473,39 @@ func checkHttpMonitorExists(dynatraceApiClient *dynatrace.DynatraceApiClient, ho
...
+		return dynatraceClusterRegionName, nil
+
+		//private
+	} else if monitorLocationType == "Private" {
+		dynatraceClusterRegionName = "backplane"
+		return dynatraceClusterRegionName, nil
+	}
+
+	return "", fmt.Errorf("monitorLocationType '%s' not supported", monitorLocationType)
+}
```
mjlshen, line 507:
```suggestion
func determineDynatraceClusterRegionName(clusterRegion string, monitorLocationType string) (string, error) {
	if monitorLocationType == "PublicAndPrivate" {
		return getDynatraceEquivalentClusterRegionName(clusterRegion)
	} else if monitorLocationType == "Private" {
	  // For "Private" HCPs, we set the region to backplane because ...
		return "backplane", nil
	}

	return "", fmt.Errorf("monitorLocationType '%s' not supported", monitorLocationType)
}
```
I would prefer if we use comments to explain the context behind some logic. E.g. the function is named `determine...RegionName`, so the public case makes sense, but the private one isn't so obvious.

Alternatively, if you like the switch/case syntax better it could also be used here, e.g.:
```go
switch monitorLocationType {
  case "PublicAndPrivate":
    return getDynatraceEquivalentClusterRegionName(clusterRegion)
  case "Private":
    // For "Private" HCPs, we set the region to backplane because ...
		return "backplane", nil
  default:
    return "", fmt.Errorf("monitorLocationType '%s' not supported", monitorLocationType)
}
```

controllers/hostedcontrolplane/hostedcontrolplane_test.go
```diff
@@ -1215,3 +1215,189 @@ func TestAPIClient_DeleteDynatraceHTTPMonitorResources(t *testing.T) {
...
+			expectId:            "backplane",
+			expectError:         false,
+		},
+		{
+			name:                "Invalid region for PublicAndPrivate",
+			clusterRegion:       "invalid-region",
+			monitorLocationType: "PublicAndPrivate",
+			expectId:            "",
+			expectError:         true,
+		},
```
mjlshen, line 1320: If we have an `Invalid region for Private` test case, we would get `expectId: "backplane"` and `expectError: false` right? Is that ok (I think so, but just clarifying)?

mjlshen (commented):
Looks good overall, non-blocking comments/questions

controllers/hostedcontrolplane/hostedcontrolplane.go
hbhushan3: Added comments - 
```go
func determineDynatraceClusterRegionName(clusterRegion string, monitorLocationType string) (string, error) {
	//public
	if monitorLocationType == "PublicAndPrivate" {
		return getDynatraceEquivalentClusterRegionName(clusterRegion)
	} else if monitorLocationType == "Private" {
		/*
			For "Private" HCPs, we have one backplane location deployed per dynatrace tenant. E.g. "name": "backplanei03xyz"
			"backplane" is returned from this function and passed to GetLocationEntityIdFromDynatrace function and this location is
			searched for in dynatrace - if strings.Contains(loc.Name, locationName) && loc.Type == "PRIVATE" && loc.Status == "ENABLED".
			Ref: https://issues.redhat.com/browse/OSD-25167
		*/
		return "backplane", nil
	}
	return "", fmt.Errorf("monitorLocationType '%s' not supported", monitorLocationType)
}
```

hbhushan3: Also added comments here https://github.com/openshift/route-monitor-operator/blob/86d2fbb61c3c6e143256b9ff3d3c604d8fc1f7a0/controllers/hostedcontrolplane/hostedcontrolplane.go#L531 to explain the functions that are being called.

controllers/hostedcontrolplane/hostedcontrolplane_test.go
hbhushan3: Added this test case. Thanks! This is okay considering we are calling `clusterRegion, err := getClusterRegion(hostedcontrolplane)` just [before](https://github.com/openshift/route-monitor-operator/blob/86d2fbb61c3c6e143256b9ff3d3c604d8fc1f7a0/controllers/hostedcontrolplane/hostedcontrolplane.go#L534) `determineDynatraceClusterRegionName` function and are returning `"backplane"` in `determineDynatraceClusterRegionName` function for Private test case.

mjlshen (approved):
/approve

controllers/hostedcontrolplane/hostedcontrolplane.go
```diff
@@ -431,6 +473,35 @@ func checkHttpMonitorExists(dynatraceApiClient *dynatrace.DynatraceApiClient, ho
...
+	}
+
+	return clusterRegion, nil
+}
+
+func determineDynatraceClusterRegionName(clusterRegion string, monitorLocationType string) (string, error) {
+	//public
+	if monitorLocationType == "PublicAndPrivate" {
+		return getDynatraceEquivalentClusterRegionName(clusterRegion)
+	} else if monitorLocationType == "Private" {
```
tnierman, line 493: Can we use the [hypershift-defined values](https://github.com/openshift/hypershift/blob/main/api/hypershift/v1alpha1/hostedcluster_types.go#L1042-L1056) instead of hardcoded strings?

pkg/dynatrace/dynatrace.go
```diff
@@ -195,28 +163,50 @@ func (dynatraceApiClient *DynatraceApiClient) GetDynatraceEquivalentClusterRegio
...
-			return loc.EntityId, nil
+
+	if locationType == "PublicAndPrivate" {
+		for _, loc := range locationResponse.Locations {
+			if loc.Name == locationName && loc.Type == "PUBLIC" && loc.CloudPlatform == "AMAZON_EC2" && loc.Status == "ENABLED" {
+				return loc.EntityId, nil
+			}
+		}
+	}
+	if locationType == "Private" {
```
tnierman, line 198: Same here - can we use the [hypershift-defined values](https://github.com/openshift/hypershift/blob/main/api/hypershift/v1alpha1/hostedcluster_types.go#L1042-L1056) instead of hard-coded strings?

hbhushan3: Thanks! Now we're using `hypershiftv1beta1.PublicAndPrivate` and `hypershiftv1beta1.Private` instead of "PublicAndPrivate" and `"Private"`.

controllers/hostedcontrolplane/hostedcontrolplane.go
hbhushan3: Thanks! Now we're using `hypershiftv1beta1.PublicAndPrivate` and `hypershiftv1beta1.Private` instead of "PublicAndPrivate" and `"Private"`.

PR #317, opened 2024-09-13
OSD-25505: Bump memory limit for RMO controller manager

typeid:
**What?**

RMO is getting oomkilled when a lot of clusters deprovision at the same time on a MC. This PR bumps the memory limit to avoid this situation.

https://issues.redhat.com/browse/OSD-25505


mjlshen (approved):
/lgtm

PR #319, opened 2024-10-09
[OSD-23845] Add cache filter to controller manager

zmird-r:
Adds necessary options to filter objects cached to reduce memory usage. This ensures that the operator controller manager does not exceed resource limits on clusters with an unusually high number of secrets or config maps. 

Ticket: https://issues.redhat.com/browse/OSD-23845

### DIFF
--- a/main.go
+++ b/main.go
@@ -28,6 +28,7 @@ import (
 	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
 	_ "k8s.io/client-go/plugin/pkg/client/auth/gcp"
 	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/cache"
 	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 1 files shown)


main.go
```diff
@@ -102,6 +104,43 @@ func main() {
 		HealthProbeBindAddress: probeAddr,
 		LeaderElection:         enableLeaderElection,
 		LeaderElectionID:       "2793210b.openshift.io",
+        Cache: cache.Options{
+            DefaultNamespaces: map[string]cache.Config{
+                config.OperatorNamespace: {},
+            },
```
mjlshen, line 110: Do we need to cache objects in the operator namespace? The per-object caches make sense to me. I'm reading through https://github.com/kubernetes-sigs/controller-runtime/blob/main/designs/cache_options.md#example-usages

zmird-r: Not really, we could have a detailed list of every single object used, I need to add a couple more I think. In this specific case, I don't think it would make a significant difference.

mjlshen (approved):
/lgtm

PR #321, opened 2024-10-11
Revert cache filters

mjlshen:
Reverts #319 and #320. Only MC's should have a cache filter for hostedcontrolplane objects, while other clusters must not have a cache filter for hostedcontrolplane objects, otherwise the operator is crashlooping.

[OSD-23845](https://issues.redhat.com//browse/OSD-23845)


mjlshen (commented):
/label tide/merge-method-squash

PR #327, opened 2024-11-07
[OSD-23845] Add cache whitelist

zmird-r:
Adds a list of resources to cache.

Ticket: https://issues.redhat.com/browse/OSD-23845


Nikokolas3270:
PR looks fine to me.
I wonder however if it could be possible to limit the memory taken by the cache rather than filtering out the objects to cache.
Proper cache management involves a proper cleanup policy. Here, from what I understood, once an object enters in the cache, it stays there forever; that's not proper cache management & that's a problem.
I will take a look at the other cache options and let you know.
But for now, let's merge this PR

PR #337, opened 2024-12-04
Removes http monitor deployment

iamkirkbater:
Removes the HTTP Monitor Deployment for new HostedControlPlanes while we fix OSD-26967.


clcollins:
/close

Obsolete

PR #342, opened 2025-01-21
config/templates: remove base replaces

jbpratt:
I have a suspicion this is causing a problem in the olm bundle
generation, versions up to 183 are still in the bundle when they
shouldn't be.

Signed-off-by: Brady Pratt <bpratt@redhat.com>


joshbranham (commented):
/lgtm

PR #343, opened 2025-01-23
[OSD-26967] Added check to verify VPC Endpoint is available 

aliceh:
https://issues.redhat.com/browse/OSD-26967

In this commit, I added a check to ensure we are deploying HTTP monitors to Dynatrace as soon as the Control Plane is Ready when the VPCendpoint is ready.


controllers/hostedcontrolplane/hostedcontrolplane_test.go
```diff
@@ -738,6 +739,11 @@ func newTestReconciler(t *testing.T, objs ...client.Object) *HostedControlPlaneR
 		t.Errorf("failed to add routev1 to scheme: %v", err)
 	}
 
+	err = avov1alpha2.AddToScheme(s)
+	if err != nil {
+		t.Errorf("unable to add avov1alpha2 scheme to test: %v", err)
+	}
+
```
tnierman, line 746: Looks like we add the new GVK to the scheme in the test file, but not in `main.go`. Does that need to happen in this PR as well?

I see the schemes used by these controllers defined [here](https://github.com/openshift/route-monitor-operator/blob/master/main.go#L61-L73) fwiw

PR #350, opened 2025-03-03
Improve the isVpcEndpointReady lookup by calling APIServer direct

jimdaga:
While testing some of the latest improvements to the `isVpcEndpointReady` logic it was found this step can cause significant delay in creating Dynatrace Synthetic monitors because it's using a cached client.

Here is an example timeline of an HCP cluster I built and how long it took the RMO operator to add my monitor:
>>>
16:26 - Cluster create starts 
16:36 - RMO notices new cluster and starts reconcile 
16:40 - Cluster in Ready state
16:41 - RMO moves onto isVpcEndpointReady() logic
17:28 - RMO finally creates the monitor, 48 minutes after the cluster entered Ready 
>>>

After some digging, I have found this to be because `mgr.GetClient` provided by controller-runtime uses a cache to improve performance by reducing the number of API calls.
https://github.com/kubernetes-sigs/controller-runtime/blob/main/pkg/cluster/cluster.go#L52-L56 

The `mgr.GetAPIReader()` function returns a non-caching client (ApiReader), which always makes direct API calls instead of using the cache.
https://github.com/kubernetes-sigs/controller-runtime/blob/main/pkg/cluster/cluster.go#L67-L70 

To keep this change safe, I have wrapped the `isVpcEndpointReady()` logic inside a `checkHttpMonitorExists()` check. We do NOT want to be making a direct API Server call on EVERY reconcile. This change makes it so we only do direct API calls when the cluster is newly created and we don't have Dynatrace monitors setup yet.


jimdaga:
Found this wasn't the problem at hand, closing PR

PR #353, opened 2025-04-15
Decrease polling duration from 10m to 2m

btoll:
This change simply reduces the polling duration timeout for the tests that rely upon it.  I've tested thoroughly and believe this to be stable.

This is part of an overall effort to help stabilize ROSA cluster operators.  Most duration timeouts are two minutes and this is normalizing that.  If a resource needs longer than that, an effort should be made to call it out as that would be an anomaly.

https://issues.redhat.com/browse/OSD-29397


ktrainor7 (approved):
/lgtm I'll be sure to pull down and test this but I trust your judgement

PR #357, opened 2025-05-06
Improve Dynatrace HTTP Monitor lookup to filter on cluster-id tag

jimdaga:


controllers/hostedcontrolplane/hostedcontrolplane.go
```diff
@@ -579,7 +562,7 @@ func (r *HostedControlPlaneReconciler) deployDynatraceHttpMonitorResources(ctx c
 	}
 	if exists {
 		log.Info(fmt.Sprintf("HTTP monitor label found. Skipping creating a monitor for %s", monitorName))
-		// return nil
+		return nil
```
joshbranham, line 565: This was already fixed on master right, meaning this is just because of a rebase being needed (not a big deal)?

controllers/hostedcontrolplane/hostedcontrolplane_test.go
```diff
@@ -759,10 +760,18 @@ func setupMockServer(handlerFunc http.HandlerFunc) string {
...
-	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-		w.WriteHeader(statusCode)
-		w.Write([]byte(responseBody)) // nolint:errcheck
-	})
+	return func(w http.ResponseWriter, r *http.Request) {
+		switch {
+		// Mock GET synthetic monitor response
+		case r.Method == http.MethodGet && r.URL.Path == "/synthetic/monitors/" && r.URL.RawQuery == "tag=cluster-id:mock-cluster-id":
+			w.WriteHeader(http.StatusOK)
+			_, _ = w.Write([]byte(`{"monitors":[{"entityId":"mock-monitor-id"}]}`))
```
joshbranham, line 768: Is there a reason you can't just pass this byte slice as `responseBody` from your test call?

pkg/dynatrace/dynatrace.go
```diff
@@ -151,6 +151,31 @@ func (dynatraceApiClient *DynatraceApiClient) MakeRequest(method, path string, r
 	return dynatraceApiClient.httpClient.Do(req)
 }
 
+func (dynatraceApiClient *DynatraceApiClient) GetListDynatraceHttpMonitors(clusterId string) (ExistsHttpMonitorInDynatraceResponse, error) {
```
joshbranham, line 154: This might be more obvious if named `GetHttpMonitorsForCluster`. Which also calls out that the functions in this file are a bit redundantely named. The package and type having `dynatrace` in the name means it would be simpler to have `dynatraceClient.GetThing()` versus `dynatrace.GetDynatraceThing()` but no need to fix that in this PR.

pkg/dynatrace/dynatrace_test.go
```diff
@@ -13,11 +14,20 @@ func setupMockServer(handlerFunc http.HandlerFunc) string {
...
-	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-		w.WriteHeader(statusCode)
-		w.Write([]byte(responseBody)) // nolint:errcheck
-	})
+	return func(w http.ResponseWriter, r *http.Request) {
+		switch {
+		// Mock GET synthetic monitor response
+		case r.Method == http.MethodGet && r.URL.Path == "/synthetic/monitors/" && r.URL.RawQuery == "tag=cluster-id:mock-cluster-id":
+			w.WriteHeader(http.StatusOK)
+			_, _ = w.Write([]byte(`{"monitors":[{"entityId":"mock-monitor-id"}]}`))
```
joshbranham, line 22: Same comment as the other tests on passing this byte slice as your response

pkg/dynatrace/dynatrace.go
```diff
@@ -247,48 +272,51 @@ func (dynatraceApiClient *DynatraceApiClient) CreateDynatraceHttpMonitor(monitor
...
 	if err != nil {
 		return err
 	}
-	defer resp.Body.Close()
 
-	//monitor already deleted
-	if resp.StatusCode == http.StatusNotFound {
-		return nil
-	}
+	for _, monitor := range existsHttpMonitorResponse.Monitors {
```
joshbranham, line 308: It doesn't look like a logger exists here, but this would be a good spot to `if len(monitors) == 0 { log.Info("No monitors found for cluster") }`

pkg/dynatrace/dynatrace.go
joshbranham, line 154: You could also have this return a pointer `*ExistsHttpMonitorInDynatraceResponse` then in your error handling, return `nil` when returning an error. It's semantic and you can leave it as is, but that's another common pattern.

pkg/dynatrace/dynatrace.go
```diff
@@ -151,6 +151,31 @@ func (dynatraceApiClient *DynatraceApiClient) MakeRequest(method, path string, r
...
+
+	if err := json.NewDecoder(resp.Body).Decode(&existsHttpMonitorResponse); err != nil {
+		return existsHttpMonitorResponse, fmt.Errorf("error parsing JSON: %w", err)
+	}
+
+	if len(existsHttpMonitorResponse.Monitors) == 0 {
+		return existsHttpMonitorResponse, nil
+	}
+
+	return existsHttpMonitorResponse, nil
```
joshbranham, line 176: Aren't these the same things? You marshall on line 168 the json into the type, then on 172 return that object if the length is 0, and then on line 176 you also return that same object if > 0

controllers/hostedcontrolplane/hostedcontrolplane.go
jimdaga: Correct, I merged my other branch into this branch (as this branch was opened first).
I needed the fix so I could do some manual testing.

controllers/hostedcontrolplane/hostedcontrolplane_test.go
jimdaga: In the way the current code is wrote without the use of interfaces, I have to handle some intermediary http responses for downstream http calls that expect different responses from what this test is expecting. We can consider a interface refactor a future TODO.

pkg/dynatrace/dynatrace.go
jimdaga: Agree, that is a good point WRT the naming. 
Resolving comment without fix as almost every function has the same issue and I want to keep this PR scoped right now

pkg/dynatrace/dynatrace.go
jimdaga: Do you think I should wire up `logr.Logger` for this purpose? I have a couple of places that would need that passed though.

joshbranham: Not as part of this PR no, but would be good in the future

pkg/dynatrace/dynatrace.go
jimdaga: This has been fixed

pkg/dynatrace/dynatrace.go
jimdaga: @joshbranham  do you mind checking that I updated this correctly as you suggested?

controllers/hostedcontrolplane/hostedcontrolplane.go
```diff
@@ -512,15 +491,13 @@ func GetAPIServerHostname(hostedcontrolplane *hypershiftv1beta1.HostedControlPla
...
-		}
+	clusterId := hostedcontrolplane.Spec.ClusterID
+	exists, err := dynatraceApiClient.ExistsHttpMonitorInDynatrace(clusterId)
+	if err != nil {
+		return false, fmt.Errorf("failed calling ExistsHttpMonitorInDynatrace [clusterId:%s]: %v", clusterId, err)
+	}
+	if exists {
+		return true, nil
 	}
 	return false, nil
```
tnierman, line 502:
```suggestion
	return exists, nil
```

pkg/dynatrace/dynatrace.go
```diff
@@ -151,6 +151,27 @@ func (dynatraceApiClient *DynatraceApiClient) MakeRequest(method, path string, r
 	return dynatraceApiClient.httpClient.Do(req)
 }
 
+func (dynatraceApiClient *DynatraceApiClient) GetListDynatraceHttpMonitors(clusterId string) (*ExistsHttpMonitorInDynatraceResponse, error) {
```
tnierman, line 154: The naming conventions in this package are generally not the ones I'd use, but would it make sense to simplify this function's name to something like the following?
```suggestion
func (dynatraceApiClient *DynatraceApiClient) GetDynatraceHttpMonitors(clusterId string) (*ExistsHttpMonitorInDynatraceResponse, error) {
```

pkg/dynatrace/dynatrace.go
```diff
@@ -247,48 +268,51 @@ func (dynatraceApiClient *DynatraceApiClient) CreateDynatraceHttpMonitor(monitor
...
-			return true, nil
+	case countMonitors == 0:
+		return false, nil
+
+	case countMonitors > 1:
+		for i := 1; i < countMonitors; i++ {
+			if err := dynatraceApiClient.DeleteDynatraceMonitorByCluserId(clusterId); err != nil {
+				return false, fmt.Errorf("failed to delete monitors for cluster id %s: %w", clusterId, err)
+			}
 		}
```
tnierman, line 291: Would it make more sense to split the deletion logic out of this function?

I personally wouldn't expect calling a function named `ExistsHTTPMonitorInDynatrace()` to have the side-effect of deleting extraneous monitors. If there's a technical reason why we must include it in this function, can we be sure to document it with comments?

PR #379, opened 2025-07-16
SREP-231 - Return error if 'openshift-ingress-operator/default' IngressController cannot be retrieved

tnierman:
https://issues.redhat.com/browse/SREP-231

---

Avoid obfuscating errors when retrieving the `openshift-ingress-operator/default` IngressController to ensure we're scheduling the `blackbox-exporter` pods appropriately


tnierman:
It's looking like we'll actually need to completely re-write the tests for this package in order to merge this code

PR #406, opened 2025-09-05
[SREP-833] Add ConfigMap support

dustman9000:
Add ConfigMap support
- Add ConfigMap reading functionality for probe-api-url configuration
- Update OLM template for MC-only configuration deployment

This ensures the operator can read probe API URL from the route-monitor-operator-config ConfigMap as specified in the OLM deployment template, while maintaining backward compatibility with command-line flag configuration.

### DIFF
--- a/main.go
+++ b/main.go
@@ -100,6 +100,17 @@ func main() {
 
 	ctrl.SetLogger(zap.New(zap.UseFlagOptions(&opts)))
 
+	// Try to read probe API URL from ConfigMap, fallback to command-line flag
+	configMapProbeAPIURL, err := getProbeAPIURLFromConfigMap()
+	if err != nil {
+		setupLog.Error(err, "failed to read probe API URL from ConfigMap, using command-line flag")
+	} else if configMapProbeAPIURL != "" {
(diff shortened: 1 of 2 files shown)


main.go
```diff
@@ -100,6 +100,17 @@ func main() {
...
 
+	// Try to read probe API URL from ConfigMap, fallback to command-line flag
+	configMapProbeAPIURL, err := getProbeAPIURLFromConfigMap()
+	if err != nil {
+		setupLog.Error(err, "failed to read probe API URL from ConfigMap, using command-line flag")
+	} else if configMapProbeAPIURL != "" {
+		setupLog.Info("Using probe API URL from ConfigMap", "probeAPIURL", configMapProbeAPIURL)
+		probeAPIURL = configMapProbeAPIURL
+	} else {
+		setupLog.Info("No probe API URL found in ConfigMap, using command-line flag", "probeAPIURL", probeAPIURL)
```
tnierman, line 111: This is the expected case for non-MC clusters, correct? Rather than `using command-line flag`, should the log state something along the lines of:

```suggestion
		setupLog.Info("No probe API URL ConfigMap found")
```

My concern is that someone not familiar with these changes could interpret this as an error

PR #417, opened 2025-09-08
[SREP-1297] Add additional routemonitor for the external FQDN.

bergmannf:
This is currently needed for [SREP-1297](https://issues.redhat.com//browse/SREP-1297) to ensure better monitoring coverage, as Dynatrace alerts are just warnings and will be replaced by RHOBS.

It adds an additional RouteMonitor for the kube-apiserver route that routes to the external FQDN, which will implicitly monitor the router-pods of the respective HCP.


controllers/hostedcontrolplane/hostedcontrolplane.go
```diff
@@ -253,6 +261,89 @@ func (r *HostedControlPlaneReconciler) isVpcEndpointReady(ctx context.Context, h
...
+		Namespace: hostedcontrolplane.Namespace,
+		Name:      "kube-apiserver",
+	}, &apiserverRoute)
+	if err != nil {
+		log.Error(err, "failed to retrieve 'kube-apiserver' route - can not setup routemonitor")
+		return err
+	}
+	apiserverServiceName := apiserverRoute.Spec.To.Name
+	if apiserverServiceName == "" {
+		return fmt.Errorf("kube-apiserver route is not referring to a service in 'Spec.To.Name'")
```
jimdaga, line 279: For a cluster with it's API Listing set to `internal` there will not be a `kube-apiserver` route by default (it's named `kube-apiserver-internal`). Can we maybe check if a cluster is internal before calling `deployExternalMonitoringObjects()` and skip the check when internal?

dustman9000:
Closing as not necessary due to multiple efforts for external synthetic approaches (Dynatrace + RHOBS.next).

PR #428, opened 2025-10-02
[SREP-1909] Fix ConfigMap env vars and add DeleteProbe debugging

dustman9000:
Configure manager deployment to read PROBE_API_URL and OIDC credentials from the route-monitor-operator-config ConfigMap. Add detailed logging to DeleteProbe method to diagnose why probe termination PATCH requests are not being sent.


clcollins (approved):
/lgtm

PR #430, opened 2025-10-02
SREP-1919 - Deploy internal monitoring objects once both kube-apiserver reachable and vpce is ready

tnierman:
https://issues.redhat.com/browse/SREP-1919

---

Re-organizes the hostedcontrolplane controller's reconcile logic slightly to check the readiness of the vpc endpoint before deploying internal monitoring objects. 

Also slight refactor to hcpReady in order to more clearly distinguish between an error checking the readiness of the kube-apiserver, and a readiness check failure.

Finally, the minimum age needed before we setup probing for an HCP regardless of kube-apiserver health is increased from 1h -> 3h

### DIFF
--- a/controllers/hostedcontrolplane/hostedcontrolplane.go
+++ b/controllers/hostedcontrolplane/hostedcontrolplane.go
@@ -37,8 +37,6 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 	ctrl "sigs.k8s.io/controller-runtime"
 
-	avov1alpha2 "github.com/openshift/aws-vpce-operator/api/v1alpha2"
-
 	"sigs.k8s.io/controller-runtime/pkg/builder"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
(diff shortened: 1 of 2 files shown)


controllers/hostedcontrolplane/hostedcontrolplane.go
```diff
@@ -222,39 +226,6 @@ func (r *HostedControlPlaneReconciler) Reconcile(ctx context.Context, req ctrl.R
 	return ctrl.Result{}, err
 }
 
-// isVpcEndpointReady checks if the VPC Endpoint associated with the HostedControlPlane is ready.
-func (r *HostedControlPlaneReconciler) isVpcEndpointReady(ctx context.Context, hostedcontrolplane *hypershiftv1beta1.HostedControlPlane) (bool, error) {
```
dustman9000, line 226: Instead of moving this entire function to a different file can't you just rename to `IsVpcEndpointReady` so it can be used by the other file?

tnierman: This isn't actually a functional change - more of an organizational one since the `isVpcEndpointReady()` function seemed more appropriate in the healthcheck.go file :slightly_smiling_face:

PR #440, opened 2025-11-04
[Srep 1632] - Add full E2E testing for RMO

Mhodesty:
Uses Rhobs-Agent and Rhobs-API repos to run a full e2e test. 
Works by utilizing local repos for testing. I have also confirmed that if we break the rhobs repos intentionally, the tests fail.


test/e2e/api_manager.go
```diff
@@ -0,0 +1,364 @@
...
+func NewRealAPIManager() *RealAPIManager {
+	ctx, cancel := context.WithCancel(context.Background())
+
+	// Get API path - must be set via environment variable
+	// The API must be available as source code to build the binary
+	apiPath := os.Getenv("RHOBS_SYNTHETICS_API_PATH")
+	var needsCopy bool = false
+
+	// If using module cache, copy to a writable temp directory
+	if needsCopy {
```
clcollins: How is this ever `true`?

Makefile
```diff
@@ -77,6 +77,35 @@ vet:
...
+		echo "Auto-detected RHOBS Synthetics API at: $$API_PATH"; \
+	fi; \
+	if [ -z "$$RHOBS_SYNTHETICS_AGENT_PATH" ] && [ -d "../rhobs-synthetics-agent" ]; then \
+		AGENT_PATH="$$(cd ../rhobs-synthetics-agent && pwd)"; \
+		echo "Auto-detected RHOBS Synthetics Agent at: $$AGENT_PATH"; \
+	fi; \
+	cd test/e2e && \
+		RHOBS_SYNTHETICS_API_PATH="$$API_PATH" \
+		RHOBS_SYNTHETICS_AGENT_PATH="$$AGENT_PATH" \
+		./run-e2e.sh
```
clcollins: Attempting to just run this directly from the command line failed:

```
RHOBS_SYNTHETICS_API_PATH=~/Projects/github.com/rhobs/rhobs-synthetics-api/ RHOBS_SYNTHETICS_AGENT_PATH=~/Projects/github.com/rhobs/rhobs-synthetics-agent/ ./run-e2e.sh
==============================================
Route Monitor Operator - Full E2E Test
==============================================

✓ Using local RHOBS Synthetics API from: /home/chcollin/Projects/github.com/rhobs/rhobs-synthetics-api/
✓ Using local RHOBS Synthetics Agent from: /home/chcollin/Projects/github.com/rhobs/rhobs-synthetics-agent/

Running full integration test...

=== RUN   TestFullStackIntegration
    full_integration_test.go:54: Mock Dynatrace server started at http://127.0.0.1:34877
    full_integration_test.go:58: Mock probe target server started at http://127.0.0.1:45717
Building rhobs-synthetics-api...
Makefile:44: *** target pattern contains no '%'.  Stop.
    full_integration_test.go:65: Failed to start API server: failed to build API: failed to build API: exit status 2
```

Is that "Makefile" reference the rhobs-synthetics-api Makefile?  The local RMO one should no longer be in the pipeline this way, I think.

clcollins: This ended up being a lack of golangci-lint installed in my local BIN (the RHOBS API Makefile could probably use some improvement), so this was I suppose a good catch via e2e, kinda.

The error message is good enough I think, but the multiple wrapped "failed to build" errors, and the exist status being returned made it a bit confusing to figure out, especially with the actual error message a line above that.

pkg/util/reconcile/reconcile_wrapper_test.go
```diff
@@ -23,7 +23,6 @@ var _ = Describe("ReconcileWrapper", func() {
 
 			ctrlResult := result.Convert()
 
-			Expect(ctrlResult.Requeue).To(BeTrue())
```
clcollins: What is the context with these being removed?

test/e2e/agent_manager.go
```diff
@@ -0,0 +1,283 @@
...
+			"error", "ERROR", "Error:",
+			"Probe .* processed",
+		}) {
+			// Skip overly verbose source info in JSON logs
+			if strings.Contains(line, `"source":`) {
+				// Parse JSON and extract just time, level, and msg
+				var logEntry map[string]interface{}
+				if err := json.Unmarshal([]byte(line), &logEntry); err == nil {
+					if msg, ok := logEntry["msg"].(string); ok {
+						level := "INFO"
```
clcollins: Should this be configurable for troubleshooting, or at least set as a constant at the top?

test/e2e/agent_manager.go
```diff
@@ -0,0 +1,283 @@
...
+		agentPath: agentPath,
+		apiURL:    apiURL,
+		stopChan:  make(chan struct{}),
+		ctx:       ctx,
+		cancel:    cancel,
+	}
+}
+
+// Start builds and starts the agent
+func (m *AgentManager) Start() error {
```
clcollins: There's  a lot of overlap in the methods for the agent and the api.  The types share significant similarities too.  Could they be a bit more DRY, implementing a shared interface for a lot of the bootstrapping/start/stop work?

test/e2e/agent_manager.go
```diff
@@ -0,0 +1,283 @@
...
+	m.started = false
+	close(m.stopChan)
+
+	return nil
+}
+
+// isPortAvailable checks if a port is available
+func isPortAvailable(port int) bool {
+	address := fmt.Sprintf(":%d", port)
+	listener, err := net.Listen("tcp", address)
```
clcollins: > For TCP networks, if the host in the address parameter is empty or a literal unspecified IP address, Listen listens on all available unicast and anycast IP addresses of the local system.

This opens that port on the host to the entire network.  I think we should default to localhost/127.0.0.1 unless we're _explicitly_ clear to anyone running this test that they're opening a pubic port.

clcollins (commented):
I made a pass with some comments/suggestions here.  I think it would be good to sit down and run though the bits doing the actual tests so I understand how it's working - I'm having a hard time following the real and fake bits, but I think that would be the case for anyone, tbh, and not any issue with your design or code.

I think you're already capturing the output logs from RMO/Agent/API via the channel/streams, is that correct?  It might clarify (and probably help with development, too, actually) if a flag could be specified to print them as they occur with something like:

[COMPONENT] [ORIGINAL LOG MSG] 

So you could see what each one is doing as the probe is created, live.  Maybe that would be easy to implement a `VERBOSE=true` with the Make target?

I'll follow up with you next week to find a time to sit down and run through the rest of this.

Thanks for the work here!  It was really great seeing it work, and the orchestration with the different repos is pretty slick. :)

pkg/util/reconcile/reconcile_wrapper_test.go
Mhodesty: This was deprecated: https://github.com/kubernetes-sigs/controller-runtime/blob/main/pkg/reconcile/reconcile.go#L30

test/e2e/agent_manager.go
Mhodesty: I moved a lot of this logic out now and created the process_manager.go file for it.

go.mod
```diff
@@ -1,90 +1,101 @@
 module github.com/openshift/route-monitor-operator
 
-go 1.23.0
+go 1.24.1
```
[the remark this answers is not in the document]
Mhodesty: I have no problem putting this back down to 1.23.0, but it looks like the most current version of go is 1.25.4.
https://go.dev/dl/

Mhodesty:
> Quick question on pkg/util/reconcile/reconcile_wrapper_test.go - looks like some mock generation directives were removed. Was that intentional or should those be restored?

@dustman9000 https://github.com/kubernetes-sigs/controller-runtime/blob/main/pkg/reconcile/reconcile.go#L30

I think this is what you're referring to on that file. That particular feature was deprecated upstream.

test/e2e/full_integration_test.go
```diff
@@ -0,0 +1,354 @@
+//go:build e2e
+// +build e2e
```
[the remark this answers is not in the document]
Mhodesty: The intention of these tests is for local e2e testing so I don't think it needs to be run by the pipeline.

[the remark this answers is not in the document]
ritmun: agreed. we spend one cluster on e2e testing prior to promotion anyway, why not automate extra coverage. 

It will come with some work because looks like you need to provide more env vars and/or secrets. But worth backlog.

